PR #78, opened 2011-08-12
Would it be possible to add jqueryui themes, this is one of the best:

graingert:
Added the aristo theme from https://github.com/taitems/Aristo-jQuery-UI-Theme/tree/master/css/Aristo


graingert:
@thomasdavis not a very good way of handling someone's first ever pull request

PR #1400, opened 2013-06-14
masonry v3.0.0

desandro:
downloaded from http://masonry.desandro.com

v3 is componentized, so masonry.js doesn't actually do much. I have included the packaged versions masonry.pkgd.js and masonry.pkgd.min.js, which can be included in any page.


mattstein:
I've hit this issue with a percentage-based layout and can tack on `style[ prop ] != 'NaN%'` to avoid the resulting error in IE8, but the cards are absolute-positioned on top of each other.

PR #2178, opened 2013-11-21
Create respond-proxy.html

termopro:
This file is used by Respond.js to access CSS files hosted on cloudflare.com CDN in situations when both CSS and Respond.js are served from cloudflare. Without it, Respond.js won't work because of cross-domain communication restrictions.

This is described by Respond.js author below:
https://github.com/scottjehl/Respond/tree/cross-domain#cdnx-domain-setup


terinjokes:
@skibulk the subdomain is not allowed to serve html documents.

PR #3883, opened 2014-09-30
[author] Update angular-route-segment to 1.3.3

artch:
https://github.com/artch/angular-route-segment/releases/tag/v1.3.3

3 commits because of 3 files edited right in GitHub web UI.


artch:
@PeterDaveHello In fact, git history is not the crux here. Even shallow clone will download around 7 Gb of data, because it's just /ajax/libs folder size.

P.S. Sorry for necroposting, I am back to update this library to 1.4.0 and it instantly reminded my all the inconvenience of this git process.

PeterDaveHello:
So I'll respectfully suggest you to add auto-update config for it, and we'll separate the real files to another place in future.

PR #4064, opened 2014-12-02
Update npmFileMap to the new built package for JsSIP

romgain:
JsSIP now builds a package file that does not contain the version number. This pull request should pick up the new versions correctly.
Note however that the new built package will be called "jssip.js" and not "jssip.min.js". I hope this is not an issue.

### DIFF
--- a/ajax/libs/jssip/package.json
+++ b/ajax/libs/jssip/package.json
@@ -6,12 +6,12 @@
     {
       "basePath": "/builds/",
       "files": [
-        "*.min.js"
+        "jssip.js"
       ]
     }
   ],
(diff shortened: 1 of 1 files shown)


drewfreyling:
@PeterDaveHello I don't see 0.6 in the repo?

PeterDaveHello:
Don't know of I remembered it wrong or it was remove rapidly, there is document for 0.6.0 here:
http://jssip.net/documentation/0.6.x/overview/

Because this commit changed the filename field, so it's real files is needed.

drewfreyling:
Closed due to no activity

PR #4070, opened 2014-12-03
Create payment-ext

hmwen:


PeterDaveHello:
@hmwen  That's not the way we add a lib, please read the document first, thanks.

PR #4128, opened 2014-12-23
Updated library epoch (http://fastly.github.io/epoch/)

lstep:
Using the download from http://fastly.github.io/epoch/download/epoch.0.6.0.zip


spotwilliams:
Sorry, but the page is 404!

PeterDaveHello:
@spotwilliams it's the old link. Try the latest version here: https://cdnjs.com/libraries/epoch/

Issue #3873, opened 2014-09-26
Semantic-ui not updated


PeterDaveHello:
semantic-ui is v0.19.3 here https://www.npmjs.org/package/semantic-ui
but seems our npm auto-update didn't update it for a while

PeterDaveHello:
no problem now

PR #4151, opened 2015-01-01
fixed auto-update bug

jdanyow:
Auto-update doesn't seem to be working.  Found a similar issue here:  https://github.com/cdnjs/cdnjs/issues/3873

The commit for this issue renamed the "repositories" property to "repository".

### DIFF
--- a/ajax/libs/breezejs/package.json
+++ b/ajax/libs/breezejs/package.json
@@ -26,7 +26,7 @@
       "web": "http://www.getbreezenow.com/"
     }
   ],
-  "repositories": [
+  "repository": [
     {
       "type": "git",
       "url": "https://github.com/Breeze/breeze.js.git"


PeterDaveHello:
oh cool, thanks!

jdanyow:
@PeterDaveHello thanks for merging this- is auto update for breezejs fixed now?  It looked like you had manually added the 1.5.0 and 1.5.1 versions here:
https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/breezejs

PeterDaveHello:
I will always audit the update result and commit by myself.

jdanyow:
I thought it was fully automated based on what's written here:
https://github.com/cdnjs/cdnjs#enabling-npm-auto-update

> cdnjs automatically updates libraries that are known to be hosted on npm e.g., Lodash. This auto-update script runs every 15 minutes.

Is fully automatic update only for select libraries or is there always a manual review+merge step?

Reason I ask is when I made the initial commit to enable auto-update I waited for the update to occur and after about an hour I assumed something was wrong with the package.json and went digging which ultimately resulted in this pull-request.  I'm still unclear whether auto-update is setup correctly because the cdn doesn't contain all of the files I thought would be included, for example `breeze.base.min.js` is missing even though `*.min.js` is specified in the `npmFileMap.files` array.

PeterDaveHello:
Yeah, used to be fully automatic, but it's not a good method, we found that there will be some problems, most the problems come from the upstream's changing on the file/directory, so I think all the commit should wait for a human to audit, another open source cdn - jsdelivr is doing the some thing, hope you can understand.

About the auto-update config, I think you are right, I'll check it.

PeterDaveHello:
I don't think it's missing ...
https://cdnjs.com/libraries/breezejs

jdanyow:
Sorry- I must have missed it before.  Very strange.   Thanks for taking the time to help me with this!

PeterDaveHello:
No problem man

PeterDaveHello:
npm auto-update of semantic-ui now works well without change "repositories" to "repository" I'm not sure if this is a real bug.

jdanyow:
ok makes sense- didn't seem related to an automated process that pulls from npm.

Noticed that breezejs 1.5.2 was added to cdnjs.  Thank you!

PR #4152, opened 2015-01-02
add jquery-snowfall v1.7.4 fixed

AppoPro:
now i think is ok.


AppoPro:
Something wrong i try to add https://github.com/loktar00/JQuery-Snowfall

PeterDaveHello:
You should add its real file(s) as the first commit of this lib, thanks.

PR #4153, opened 2015-01-03
[author][new]my first library.

kunass2:
This is my first ilbrary. And i would like it to be in cdnjs.


PeterDaveHello:
Please merge/squash all these commits into only one commit as the document said, thank you.

Issue #3638, opened 2014-07-24
No more manual pull request for lib updating


thomasdavis:
Hey guys,

We are going to try switch the project to be fully automated. This will take a bit of time and will be a bit rocky at the beginning but will save every bodies time.

We currently have two auto update systems, the old one which uses NPM and will remain active until we transfer it. And a new one that uses git tags. The new one is what we will use by default but it doesn't support NPM just yet so we can only add new auto update scripts for git. We can also add the old style NPM ones I will just transfer them at a later date.

``` json
  "autoupdate": {
    "source": "git",
    "target": "git://github.com/jashkenas/underscore.git",
    "basePath": "",
    "files": [
      "underscore-min.js",
      "underscore-min.map",
      "underscore.js"
    ]
  },
```

During this process, I think it's best if only I merge PR's to make sure we don't cause too many problems at once. I still need to write up documentation. So from this point on don't merge any new libraries into cdnjs.  @drewfreyling @PeterDaveHello @ryankirkman

JamesMGreene:
@PeterDaveHello: I see you [manually?] added ZeroClipboard `v2.2.0`. Any info on its failure to auto-update? I'll likely be releasing another version some time this week.

PeterDaveHello:
Yes, because of I'll always audit all the updated results, so you'll see my commit, we don't use a bot to commit it at this time, I don't think we have problem here, maybe you can ping me here when you're going to release the it, then I'll check if everything okay.

dlid:
Hi, I added the tag 0.8 of https://github.com/dlid/jquery-valjs/releases but I can't see it on cdnjs yet. Is there something else I must do for the auto update to work?

PeterDaveHello:
Sorry, I am too busy these days, I'll audit all the updates results before commit it, added in b9b2efcd60704f1d2ddf0e3c3656fb725fa4f47f now.

Jakobud:
I think I am missing something here.... How exactly do we add or update libraries to cdnjs now? Do we not do pull requests anymore?

The readme says "no more manual submissions" but then it says to do pull requests to add or update libraries. Which is it?

Really confused...

PeterDaveHello:
PR for adding a whole new lib or the libs can't not support/be supported by current auto-update, no manual submissions for updating libs. We are working for the new mechanism to update libs already.

ethanpil:
Maybe I'm just being dense... but I am missing something.

[...]

PR #4155, opened 2015-01-03
[author][new] ValJS V0.7

dlid:
First version of ValJS
Minified source: https://github.com/dlid/jquery-valjs/blob/master/jquery.valjs.min.js
Uncompressed source: https://github.com/dlid/jquery-valjs/blob/master/jquery.valjs.js


PeterDaveHello:
Hi @dlid, 

Is it possible to add auto-update config in it? We hope all the lib can be easily updated in future, you can use npm or git auto-update as readme and #3638 said, thanks.

dlid:
Hi, I can absolutely do that.

Do I understand it correctly that cdnjs will use git tags to keep track of the versions? So if I want my 0.7 version to be available on cdnjs I create the 0.7 tag, and for future versions when I create a new tag - it will be picked up automatically by cdnjs?

Thanks.
/David

PeterDaveHello:
Yes, you are right! So please add git auto-update config for it, thanks a lot.

dlid:
I added the autoupdate in https://github.com/dlid/jquery-valjs-cdnjs/blob/master/ajax/libs/valjs/package.json now so I hope it's correct. I also created the 0.7.3 tag in the source rep and updated the files in my cdnjs-fork to the latest files.

PeterDaveHello:
Would you please merge these two commits into only one commit? thanks.

dlid:
I've having issues merging the commits, I'm no github ninja unfortunately. Could I drop my fork and make a new pull request instead?

dlid:
Closing this request, Github is not friends with me. I'll try again with a separate request

PR #4159, opened 2015-01-04
[author][new] jquery.scregal v1.0

kunass2:
This is my first library. I would like to add it to cdnjs.


PeterDaveHello:
Would you please add atuo-update config for it as readme or #3638 said? thanks.

PR #4161, opened 2015-01-04
[author] dinqyjs Updated to 1.2

garrypas:
URL: https://github.com/garrypas/dinqyjs
A few new features have been added to this library which have now been committed to the repo and to npm. These are reflected in this change. The package.json file has also been updated accordingly.


PeterDaveHello:
Would you please change your commit log to include the lib's name as @drewfreyling done in this PR's title?

According to #3638 , you should also add auto-update config for it, thanks.

PR #4166, opened 2015-01-05
[new][author] ValJS V0.7.3

dlid:
Changed package.json to UTF-8 in an attempt to resolve build issue.
Source rep: https://github.com/dlid/jquery-valjs/tree/0.7.3


PeterDaveHello:
Hmmmmm ... still failed, I may take a look at your package.json to find out the problem.

dlid:
That would be great. I've run it through JSON Lint so it should be valid JSON. And the encoding was the only other thing I could think of. I also changed the ö-character of my surname to the escaped version...

PeterDaveHello:
bomb of utf ... I fixed it and commit here 66562f7981e7a3562fb217e2b18916b6d760d48d

PR #4171, opened 2015-01-05
Add angular-scroll

quanack:
Adding library for Angular Scroll located at https://github.com/oblador/angular-scroll

PR #4179, opened 2015-01-06
Added angular-scroll

quanack:
Adding library for Angular Scroll located at https://github.com/oblador/angular-scroll


ajax/libs/angular-scroll/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "type": "git",
+    "url": "https://github.com/oblador/angular-scroll.git"
+  },
+  "bugs": {
+    "url": "https://github.com/oblador/angular-scroll/issues"
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://git@github.com:oblador/angular-scroll.git",
```
PeterDaveHello: Wrong url format

quanack: OK, This is getting tiresome.  What is in the wrong format.  I tried this before and was told that the URL was in the wrong format and I changed it.  You need to get more descriptive on what's wrong.

Look at this https://github.com/cdnjs/cdnjs/pull/4171 where he is telling me it is the wrong format.

PeterDaveHello: Of course the url in line 25 is in wrong format, I think you can try it with git clone like this
`git clone git://git@github.com:oblador/angular-scroll.git`
and then you'll get the error message, I think you mixed the ssh and git protocol in the same url.

PeterDaveHello: http://git-scm.com/book/ch4-1.html

PeterDaveHello:
FYI, I added in ccef47fb62b2054fb37cd3a8238778d2448bf9ff.

PR #4180, opened 2015-01-06
[author] jquery.scregal v1.1

kunass2:


PeterDaveHello:
I think the config of auto-update will not work, and the commit log should include the lib's name.

kunass2:
May I ask, why my auto-update will not work?

PeterDaveHello:
A wildcard symbol will only match the first layer, doesn't contain any useful files in this project.

You can take uikit's package.json as an example :)

kunass2:
It means that:  "*" is wrong. Where can i take package.json example from?

PeterDaveHello:
From here please https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/uikit/package.json
And please leave other parts alone, like the url of the repo, there is no reason to change the protocol to https, git protocol is more efficient then https, though some firewalls may block it, but most of the users or contributors don't need to run the auto-update process, so git will be better here.

drewfreyling:
Closing because @kunass2 added #4183

Issue #3650, opened 2014-07-25
Sortable vs sortable causes issues on case insensitive filesystems


zpao:
Simply checking out cdnjs results in a file being modified because you have https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/Sortable and https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/sortable. It's not just my fs though, it looks like you own site can't handle this (search for "sortable" and you see the summary for `Sortable` but view the page and it's `sortable`).

PeterDaveHello:
See https://github.com/cdnjs/cdnjs/blob/master/sparseCheckout.md, will help.

gordonwoodhull:
New link: https://github.com/cdnjs/cdnjs/blob/master/documents/sparseCheckout.md

PeterDaveHello:
Thanks @gordonwoodhull

NGsolution:
Sortable

xtuc:
is there anything actionable here? we won't modify existing files because that would be a breaking change.

MattIPv4:
Closing as this isn't actionable, but have pinned for visibility as this will be a forever issue with this repo (and other related cdnjs repos).

japui-coder:
Sortable vs sortable causes issues on case insensitive filesystems
 #3650

PR #4183, opened 2015-01-07
[author] jquery.scregal v1.1

kunass2:
with good auto-update


PeterDaveHello:
What's the reason why to change it's main file jquery.scregal.min.js to jquery.scregal.js? minified version is preferred.

PeterDaveHello:
And I have another question, don't you need its css file?

kunass2:
Ok, i see. I know that about minified version. I change it by next pull request. And I add css file. I think i change the structure by adding two directories: css/ and js/. Is it good? Can i upload img/ with a very small images?

PeterDaveHello:
Yes, of course you can.

kunass2:
I have a question. Why every time when i fork cdnjs repository to my account, and then clone it to my computer, there is some changes ready to commot them. Why they always appear, although i changed nothing? Because of this, I cannot rebase my local repo on yours current version of cdnjs. Hence i cannot create a correct pull request. The file that I keep seeing changed is from Sortable library. Package json deleted, and created. Why?

PeterDaveHello:
#3650

kunass2:
So far, it is not resolved:-) ok, we still waiting...

PR #4188, opened 2015-01-09
[new] Added magic.css animations

miniMAC:
CSS3 Animations with special effects


PeterDaveHello:
Please read our document for what you need to send a PR, thanks.

PR #4194, opened 2015-01-10
[author] [new] Added maquette

johan-gorter:
Developed at https://github.com/johan-gorter/maquette


PeterDaveHello:
Excuse me, there are 3 files (css-transitions.min.js  maquette.min.js  velocity-transitions.min.js) under dist folder, but you only add one, is this correct?

johan-gorter:
The other files are nog yet important. It is ok if these files only appear in future versions.

PeterDaveHello:
Okay.

PR #4197, opened 2015-01-12
[author] [new] Add the Aegis Framework

kumarharsh:


PeterDaveHello:
Wrong auto-update config here ... there is no fonts folder under dist ...

kumarharsh:
@PeterDaveHello Is it possible to supply two basePaths? Or do I specify the basePath as `.` and then reference the css, and fonts explicitly?

PeterDaveHello:
Current, no, but I remember that npm auto-update support it, you can consider to use npm auto-update or change the bashPath as you said.

PR #4198, opened 2015-01-12
[new] Added angular mixpanel 1.0.0

OxanaSachenkova:
Library source: https://github.com/kuhnza/angular-mixpanel 
This is the best library I could find for integrating mixpanel with angular. Would be nice to have it on cdnjs.


dovydasnavickas:
There is also angulartics. You might want to check it out:
http://luisfarzati.github.io/angulartics/


Regards,
Dovydas Navickas
QuatroDev CEO
________________________________

PeterDaveHello:
Thanks OxanaSachenkova .

PR #4202, opened 2015-01-14
Add autoupdate for Leaflet.awesome-markers

int64ago:
Is this OK?


PeterDaveHello:
Leaflet.awesome-markers looks okay, but in KaTeX:
I don't think we need `*.less` files, and main.css & main.js are both not for release.

int64ago:
I've updated the PR :)

PeterDaveHello:
Thanks.

PR #4203, opened 2015-01-14
[new] Add angular-material

int64ago:


int64ago:
Can auto-update fetch old versions automatically if the version field is the latest version

PeterDaveHello:
Yes, would you please take a look at angular-material.min.js? I don't think it's really minfied, seems comments are duplicated.

int64ago:
It's official compressed version, I find that if comment is like this:
```
/*!
 something...
*/
```
Most of compressors can't remove comments after compression.
So I think it's better to leave minified version like this.
What do you think?

PeterDaveHello:
Okay, thanks.

PR #4208, opened 2015-01-15
[author] [new] add videojs-wavesurfer

thijstriemstra:
Files downloaded from https://github.com/collab-project/videojs-wavesurfer/releases


PeterDaveHello:
I can't find dist folder in it's package.

thijstriemstra:
Thanks for your feedback. When you download a release, eg. https://github.com/collab-project/videojs-wavesurfer/releases/download/0.9.1/videojs-wavesurfer-0.9.1.zip you'll find the `dist folder in there. I'm new to this npm packaging, but in the Python world you don't want to include the compiled files, eg. `foo.pyc`. Are you saying I should add the `dist` folder to the repository and always have a copy of the minified files in there? There's currently a grunt file that builds the dist folder's content. Ì run this locally during a release, tag the release and put up the .zip with the dist folder included. Apologies if these are dumb questions ;)

ajax/libs/videojs-wavesurfer/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "repositories": [
+    {
+      "type": "git",
+      "url": "https://github.com/collab-project/videojs-wavesurfer"
+    }
+  ],
+  "npmName": "videojs-wavesurfer",
+  "npmFileMap": [
+    {
+      "basePath": "/dist/",
```
PeterDaveHello: So this will not work properly, right?

PeterDaveHello:
If you'll add tags for each release and put dist files there, you should change npm auto-update to git atuo-update, because there is no dist in npm package, ref #3638 as git auto-update example.

drewfreyling:
Closing due to no response.

PR #4211, opened 2015-01-15
[author] [new] Add pusher-angular v0.1.4

hamchapman:

PR #4212, opened 2015-01-15
[author] [new] Add pusher-angular v0.1.4

hamchapman:


PeterDaveHello:
Don't you want to use npm as #4211 ? I prefer to use npm, it'll be faster to update a lib, what do you think?

hamchapman:
I'm not too bothered either way, but I'd rather have it update via git if possible. If it's a big advantage for you guys to have it updated via npm then I'm okay with that, but my personal preference is for git.

PeterDaveHello:
Yeah, it's a big advantage, if you are sure you'll update npm package regular, I'll suggest you to use npm update, it's more efficient for us, and check for update every hour, git auto-update doesn't, because every round of git auto-update will take much time.

hamchapman:
Okay, please could you merge in #4211 instead then please? Thanks for your help!

PeterDaveHello:
Okay, I'll be happy to, but seems you deleted the branch so that I can reopen it, would you please undo and open it again? thanks.

hamchapman:
I've force pushed the npm autoupdating version to update this pull request. Hope that's okay.

PeterDaveHello:
Thank you @hamchapman.

PR #4215, opened 2015-01-16
[new] Add ng-context-menu

int64ago:
In npm, there's also a ng-context-menu, but I think this one is better.


PeterDaveHello:
Hi @int64ago,

Would you please tell me how better is it? I prefer to use npm pacakge :)

int64ago:
The version in github(https://github.com/Wildhoney/ngContextMenu) is 0.2.3 but  in npm is still 0.1.1,
so...
Of course, It's my view...

PeterDaveHello:
Solid view man, you are right!

PR #4223, opened 2015-01-18
[new] [author] add tanner-bootstrap

TheTannerRyan:


PeterDaveHello:
Hey @TheTannerRyan, I think you should add tag in your repo, and use dist or bin instead of 1.0.0 as the release file, what do you think?

TheTannerRyan:
I have renamed the release file and updated.

PeterDaveHello:
Then please change bashPath to dist, and merge/squash these commits into only one commit, thanks!

PR #4224, opened 2015-01-18
[author] jquery.scregal v1.2

kunass2:


PeterDaveHello:
Would you please use auto-update for it as #3638 said? And you should keep or update its structure to pass the build test.

PR #4228, opened 2015-01-18

kireerik:
origins:


PeterDaveHello:
Please merge/squash them into only one commit, thanks.

kireerik:
Thank you for your feedback! Sorry for misunderstanding your "Only do one thing in one commits, don't mix different things into the same commit." guideline. It seems the "Every commits should be meaningful, don't cut one thing into multiple commits." guide line has higher priority.
I will create a new pull request soon with the requested modifications.

PeterDaveHello:
Sorry for the misunderstanding, I'll update it, thanks.

PR #4233, opened 2015-01-19

kireerik:
origins:


PeterDaveHello:
Is it possible to add auto-update config for it? thanks.

kireerik:
Yes, sorry for forget about it. I do not know it is necessarily. (It was not on the pre-flight check list.)

PR #4234, opened 2015-01-19

kireerik:
origins:


PeterDaveHello:
I found that there are 3 files under `dist` but you only added 1, please add 2 others, thanks.

kireerik:
Okay, I will add them too.

PeterDaveHello:
In fact, you can just use force push to update a PR, instead of keep opening new PRs.

kireerik:
Thank you for mentioning this! I am really sorry for creating multiple pull requests with the same issue. I am a little newbie collaborating on GitHub. Hope everything will fine for this time. ;)

PR #4238, opened 2015-01-19

kireerik:
origins:
- https://github.com/nolimits4web/Swiper/blob/master/dist/


PeterDaveHello:
Thanks.

PR #4240, opened 2015-01-20
[author][new] vault.js 0.1.1

jimmybyrum:


PeterDaveHello:
Please use only one kind of auto-update, seems you add npm + git auto-update for it, thanks.

PR #4247, opened 2015-01-20
Auto-update config for stripe/jquery.payment library.

mjswensen:
Is this sufficient to kick off the automatic updating of this library, or should I include the latest version manually as well?


PeterDaveHello:
That's enough, thanks.

PR #4249, opened 2015-01-20
[author] update textAngular to v1.3.0 format

SimeonC:
Change the include paths to include all the new paths in the v1.3.0 full and pre-releases.


PeterDaveHello:
Thanks.

PR #4253, opened 2015-01-21
Add the then-request library

ForbesLindesay:
This adds the then-request library to cdnjs.  It sets it to auto-update from npm.


PeterDaveHello:
Please add its real files as the first commit to add this lib, thanks.

PR #4254, opened 2015-01-21
[new] [author] Initial release of tanner-bootstrap theme - v1.0.0

ghost:


PeterDaveHello:
Finally, looks okay now :+1: 
But I still want to confirm that there is no source file in your git repo, only minified dist file, is this correct?
Thanks.

TheTannerRyan:
Just the minified dist file. 

Also, if I just update my repo and change the version number, will it automatically update, or do I have to clone, commit and send a push request every time?

PeterDaveHello:
If you add tag for it, yes, will be automatically updated with in 24 hours. But what's the reason why don't put the source file? If we want to check the codes or contribute, it'll be better to have it, right?

TheTannerRyan:
I just added a source file

drewfreyling:
@PeterDaveHello This is one of those where I would probably err on the side of not adding it to the repo as nobody is using it yet and no starts. But your call.

PeterDaveHello:
Sounds reasonable, what do you think @thomasdavis ?

TheTannerRyan:
For some reason, my account was deleted.

PeterDaveHello:
Hey @TheTannerRyan, what happened?

TheTannerRyan:
@PeterDaveHello I have no clue.

PeterDaveHello:
I think you should add your repo back first.

PeterDaveHello:
no response?

PeterDaveHello:
Close due to no response.

PR #4255, opened 2015-01-21
add custom minified meny js

kireerik:
This minified version includes a proper Z-index for the menu element. (I usually use this modified version.)


drewfreyling:
The minified file should be an exact copy that is only minified with no other changes.

PR #4257, opened 2015-01-22
selector-queries

jmusicc:
Library that supports the use of media queries on selectors instead of viewports.


PeterDaveHello:
Please take a look at our document for what you need to send a PR, thanks.

jmusicc:
I was having trouble editing files via the website because I was using a different PC that wouldn't install the CLI. The original repo was formatted incorrectly and the script name didnt match the master. I think that's why it threw an error during the check.   I ended up just using RawGit lol

PR #4266, opened 2015-01-26
[author] [new] Add shaka-player.

joeyparrish:


joeyparrish:
This is the compiled version of the Shaka Player library, v1.1, built from source.  The original source is at https://github.com/google/shaka-player/ .

Thanks for your consideration, and please let me know if I missed anything.

PeterDaveHello:
Thanks.

PR #4270, opened 2015-01-28
Autoupdate jQuery.dotdotdot

tomsterritt:
Wasn't sure if I needed to include latest version so I did anyway, but hopefully this will autoupdate?


PeterDaveHello:
Please use git protocol in target field, thanks.

PR #4275, opened 2015-01-29
Changed Riot's package.json (added more files)

aurri:
The next Riot version will include new files: riot+compiler.js and riot+compiler.min.js.
Added them to package.json.


PeterDaveHello:
Thanks！

PR #4279, opened 2015-01-29
Adding waypoints 3.1.0

simondoty:
The most recent version of the (previously jQuery) library Waypoints is 3.1.0. The latest version on cdnjs is 2.0.5.
I would like to add the waypoints library and shortcut libraries. 
Source: https://github.com/imakewebthings/waypoints/zipball/latest


PeterDaveHello:
sortable issue, ref #3650
Please try to not touch sortable, thanks ：)

PR #4280, opened 2015-01-29
Ekene Ijeoma adding new library ijeoma.js at v1.0.5

ekeneijeoma:
https://github.com/ekeneijeoma/ijeoma.js


PeterDaveHello:
According to #3638, would you please also add auto update-config for it? thanks.

PR #4291, opened 2015-01-31
Adding Smart-Table AngularJs Module

kingsimmy:


PeterDaveHello:
Would you please use git protocol instead of https in target field and merge/squash these 2 commits into only 1 commit? thanks!

PeterDaveHello:
No response?

PR #4293, opened 2015-01-31
autoupdate for lodash

hzoo:


PeterDaveHello:
Seems there is no lodash.compat.js in repo?

hzoo:
Oh you're right - you have to use lodash-cli for that

PeterDaveHello:
So there won't be pre-built lodash.compat.js anyway?

hzoo:
It's on https://lodash.com/ but not in the repo itself

PeterDaveHello:
hmmmm .... I think we can't not use auto-update on it because we may forget to add these additional files :(

hzoo:
Based on the patch notes it looks like there's only 4 now - the underscore, backbone, and other files are gone in 3.0

PeterDaveHello:
Sorry that I didn't get your point here, would you please explain in detail? thanks.

hzoo:
Sorry I mean't that in previous versions you would upload a lot more files.
image

But now it looks like there are only 4 main files. lodash and lodash.compat (and minified versions of both).

It looks like we need https://github.com/lodash/lodash-cli to build any other ones. https://lodash.com/custom-builds

I guess it depends on how many kinds of builds should be supported. Unless autoupdate can run commands then I guess only update `lodash.js and lodash.min.js`

PeterDaveHello:
Okay, thanks for your explain, but currently, our auto-updater can't run additional commands, so I still think we can't not using auto-update this time. Another method is to ask for help from lodash team, if they can provide pre-built files in origin repo, then it'll be easier for us, what do you think?

hzoo:
Yeah I think just upload the files manually then as I don't think they would include the files in the repo - that's probably why they have the other lodash-cli project

PeterDaveHello:
Okay, I'll subscribe the rss feed of their releases and try to update as soon as they release the new versions.

hzoo:
I can do another PR to manually update 3.0.1 - but looks like you got it

jdalton:
The modern and compat files now live in separate repos.

You can find the modern files for say 3.1.0 here:
https://raw.github.com/lodash/lodash/3.1.0/lodash.js
https://raw.github.com/lodash/lodash/3.1.0/lodash.min.js

and compatibility files here:
https://raw.github.com/lodash/lodash-compat/3.1.0/lodash.js
https://raw.github.com/lodash/lodash-compat/3.1.0/lodash.min.js

hzoo:
Oh ok thanks for the clairification

PeterDaveHello:
hmmm ... but we don't support grab files from different repos even different files, so still can not apply auto-update on it.

thomasdavis:
A little confused...

So we will have the main Lodash files but not individual or compatibility files? That should be fine for now, we could just add a new `package.json` for the other files and they could exist under cdnjs.com/libraries/lodash-compat?

jdalton:
they could live under lodash-compat or mapped to the lodash/ location:

```
https://raw.github.com/lodash/lodash/<tag>/lodash.js -> lodash.js
https://raw.github.com/lodash/lodash/<tag>/lodash.min.js -> lodash.min.js
https://raw.github.com/lodash/lodash-compat/<tag>/lodash.js -> lodash.compat.js
https://raw.github.com/lodash/lodash-compat/<tag>/lodash.min.js -> lodash.compat.min.js
```

thomasdavis:
Alright, let's keep it in one directory for now and just add lodash-compat manually until we figure out a strategy for the auto update.

PeterDaveHello:
I think open a new directory as a new lib will be easier.

PeterDaveHello:
updated in a8edd13 and added compat version in 76391d6

PR #4304, opened 2015-02-02
[author][new] Add neo-async v0.5.0

suguru03:
I have developed neo-async, I would like you to merge it. 
source: https://github.com/suguru03/neo-async


PeterDaveHello:
Hi @suguru03, thanks for your PR, I noticed that you added both npm and git auto-update config, but only one is enough, do you always update the npm package on time? If so, please leave npm config alone and remove git auto-update config, thanks. BTW, we don't use ssh protocol in git auto-update, git protocol will be fine.

suguru03:
Thanks for reply.
I fixed package.json. Would you check this commit?

PeterDaveHello:
Thanks!

PR #4307, opened 2015-02-02
jquery.isotope - add npm auto-update

desandro:
Update Isotope description, homepage, maintainers, git url

### DIFF
--- a/ajax/libs/jquery.isotope/package.json
+++ b/ajax/libs/jquery.isotope/package.json
@@ -2,16 +2,18 @@
   "name": "jquery.isotope",
   "filename": "isotope.pkgd.min.js",
   "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


desandro:
On a related note, what kind of effort would it take to move the url from `jquery.isotope` to just `isotope`? jQuery is no longer required as of v2.

drewfreyling:
That would be ok for future versions but we'd have to leave the existing versions (2.0.0, 1.5.25) under the jquery.isotope name. Looping in @PeterDaveHello to get his opinion on this too.

desandro:
Ah, thanks! I'm cool that.

PeterDaveHello:
Renaming is okay, I'm going to rebuild the test, because we had a problem on circleCI and fixed now.

PR #4312, opened 2015-02-03
[new] add Menu

kireerik:
sources: https://github.com/kireerik/Menu

Note: I have made some improvements and fixes on the original [Meny](https://github.com/hakimel/Meny) and requested there a [pull request](https://github.com/hakimel/Meny/pull/43). I also drop him a mail about my pull request, but I didn't get any response. It seems to me that @hakimel no longer deal with this project (there are other 2 pull request without any response). So as Meny is MIT licensed I decided to redistribute the project under a new name.
Should I change anything on my [repository](https://github.com/kireerik/Menu) such as renaming everything and replacing the author name to mine and adding some cretits and linking to the original author?

Thank you for your feedback and thank you for accepting my pull request if it is correct as is in advance!


PeterDaveHello:
Would you please delete the old tags in the repo?  'cause they are meny, not menu, right?

kireerik:
Of course. Do I need to change the author name in the LICENSE file as well?

PeterDaveHello:
If you are developing a forked, new lib, I think the answer is yes.

kireerik:
Thank you for your feedback!
MIT license says
> The above copyright notice and this permission notice shall be included in
all copies or substantial portions of the Software.

and I read on forums that the license file should be intact. It seems this not applies to the author part.

I am planing to contact with @drgould and @kenma9123 to merge their [pull requests](https://github.com/hakimel/Meny/pulls) into my repository. So yes I am developing a forked "new" lib.

PeterDaveHello:
Maybe to can add yourself to author field, but without remove origin authors, but I'm not good at license, just person view.

kireerik:
The same is true for me and I didn't found relevant information on how to re-distribute a MIT licensed software properly.
I would be fine if @hakimel at least given some feed back on my pull request, but no. So although my contribution is really really little to the project, I think I need to redistribute it with the modifications to be able to use it via cdnjs.
Okay I will do that, it sounds correct. I will modify it if someone points out that it is nor legal, ethical or correct with the original author.

PR #4318, opened 2015-02-05
Update ember-simple-auth to 0.7.3

NathanJang:
This package is not on npm and does not have precompiled files in its
git repository.

Files were compressed with UglifyJS.

Please advise whether or not other versions should be added.


PeterDaveHello:
Thanks, is it possible to add auto-update config for it? thanks!

NathanJang:
I don't know how to add an auto-update for it. The README says auto-update only supports packages hosted on npm and Github. However, it is not hosted on npm and does not have built files in its [git repository](https://github.com/simplabs/ember-simple-auth).

PeterDaveHello:
Okay, thank you.

NathanJang:
Thanks! Should I add older versions as well?

NathanJang:
On second thought, ember-simple-auth does have built files available for download on its tags page.

PR #4332, opened 2015-02-09
Add autoupdate for ember-simple-auth

NathanJang:
I have found a git repo with built files since #4318.


PeterDaveHello:
Thanks.

PR #4336, opened 2015-02-09
Updating Library AmMaps to 3.13.0

abishekrsrikaanth:
### AmMaps

URL: http://www.amcharts.com/javascript-maps/

The above library is free to use. Please check the section "Free even
for commercial use" on the above Library's URL

Since this library is not hosted on github, auto update has not been
configured.

Removed examples, .md, .txt files from the original library as this is
not required to be hosted on cdn.

Hopefully all is OK with this PR.

@PeterDaveHello. Sending PR again as requested. Could you please verify
the same.


PeterDaveHello:
Should come with minified files next time, thanks.

PeterDaveHello:
I found that most of the files are already minified, but some not, it's hard to do the minify easily, so forget it, it's okay like this one.

PR #4341, opened 2015-02-10
Update danialfarid-angular-file-upload

ruiaraujo:
Set up danialfarid-angular-file-upload to be auto updated.


PeterDaveHello:
Nice, thank you @ruiaraujo.

PR #4356, opened 2015-02-12
change description for l20n package

gasolin:
as discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1129965#c3

We prefer to change the package statement as "A client side localization library from Mozilla".


PeterDaveHello:
Sure.

PR #4362, opened 2015-02-16
[author] Adding flocks.js

StoneCypher:
Thanks for providing a system like this to the community.  :smile: 

This is my first time working with CDNjs, so if I botched something, I'll work on it.

There is no minified version because minifiers don't play well with JSX right now.  I'm looking for a resolution.


StoneCypher:
Fixing.

PeterDaveHello:
Is it possible to change its extension from jsx to js? I am not sure if we support file with *.jsx naming, I guess not, and I'll need you to squash these commits into only one commit as our document said, thanks!

StoneCypher:
At the time this was written it was required to be `*.jsx`.  I can look into whether that's changeable by now; there was some talk of it on the react team.  I'll squash it when I get home tonight.

Is the extension a deal-breaker?  That'd be a bummer :frowning:

StoneCypher:
Okay so I asked around some.

It is possible for it to be `*.js` now, yes.  However, what's inside is legitimately a language dialect, in the sense that coffeescript and typescript and so on are.

I could put the transformed JS version into the CDN and minify that, but having the `*.jsx` version in there would be nice for people doing development, since as far as I can tell there are no sourcemappers that make it through `*.jsx` yet.

If I can get the `*.jsx` in I would like to, but if I can't, I understand, and I can put in the transformed version instead.

I kind of feel like maybe having the `*.jsx` version in, and the `*.js` version as the "dev minified" version, might be the way to go if it works with your setup, and if you're okay with that, but I don't know your norms very well, and I'm flexible.

Please advise.

StoneCypher:
Actually on talking to the community they seem to think it's odd that I want to put the non-transformed one up at all, so, forget what I said earlier.

I will modify this PR tonight to contain the post-transform JS version instead.  Forget the `*.jsx` issue please :smile: 

Thanks for your patience with me.

PeterDaveHello:
Thanks, and if `.jsx` support is needed in same cases, we can open another issue to add it.

StoneCypher:
welp, couldn't rebase because there's two libraries in the repo that have the same name other than case, and i'm on a windows machine

so i tried to take advice how to squash without a rebase in #git on freenode, and somehow that destroyed the branch :D

will try again tomorrow; it takes six hours to check this thing out

PeterDaveHello:
Call me if you need my help :)

PR #4366, opened 2015-02-17
[author] Adding flocks.js

StoneCypher:
Re-attempt of https://github.com/cdnjs/cdnjs/pull/4362


ajax/libs/flocks/package.json
```diff
@@ -0,0 +1,92 @@
+{
+  "author": "John Haugeland",
+  "name": "flocks.js",
```
PeterDaveHello: Should be the same as its folder name here.

StoneCypher:
Yep.  Was asleep :)  Just woke up, got it

Thanks :)

PeterDaveHello:
Thanks!

StoneCypher:
Thank you very much for your help and your patience, @PeterDaveHello :hand: :fist: :facepunch: :grinning:

PeterDaveHello:
NP :+1:

StoneCypher:
@PeterDaveHello so I might not entirely understand your system.

I thought the gag with auto-update was if I pushed a new version, the CDN would pick it up on its own. Is that belief mistaken?

I ask because I thought I set auto-update up, but I pushed a small update and it hasn't been picked up yet.  Do I need to fix how I did that, or did I misunderstand what the system does?

PeterDaveHello:
You should take a look at readme and you'll know the reason why.

StoneCypher:
So I'm missing it, whatever it is.  I just read the readme three times.  I think I might be misunderstanding a piece, because I'm reading right past whatever it is.

Things I think matter:
  * There's a passing test suite, correctly hooked up to report through npm and gulp
  * I pass your pre-flight requirements
    * There's a normal and a minified `.js`
    * No CSS so CSS rules are moot
    * The lib name, npm name, main entrypoint, repo name, and directory all match verbatim
    * Your directory structure is honored
      * and matches the lib name
    * The `package.json` contains:
      * your custom `repositories` field
      * the auto-update mechanic
        * which I'm worried I got wrong but can't tell and seems right
      * the npmName matches the NPM published name
    * The `autoupdate` field
      * exists
      * matches the structure of the one in the `README`, I *think*
      *
    * The `npmFileMap`
      * exists
      * is a direct literal copy from another library with working auto-update, to reduce the potential for error

I'm not being lazy; I'm being stupid.

I was hoping you might tip me off to what stupid thing I'm doing?  :pray:

PeterDaveHello:
Hmmmm ... are you wondering what's the reason why new version didn't be picked up? 

As our document said, every new version will only be commited after I audited it, to make sure the structure and version info are correct, so there it'll delay up to 30 hours, that's the reason why you may not see it immediately.

PeterDaveHello:
BTW, I an not sure if you are talking about this one 

* 1e87b54 add flocks.js v0.14.2/0.14.4 Peter Dave Hello 28 hours ago 

Sometimes our website will delay, but the files will be on cdn with in 30 mins.

StoneCypher:
Oh!  :laughing: 

Sorry to be a nag.  I completely missed that.  I was under the impression that I had botched the technical job.  :dancers: 

Thank you very much!

PeterDaveHello:
np

PR #4371, opened 2015-02-18
make cdnjs cookie free domain

pushkeryadav:
Cookie free domain make request faster and there is no need to use cookie for static contents. I saw cdnjs adding  "__cfduid=d8ef483ca1c9e6de9814929z2a509fcb914091399382350" is less in size but we dont need this also.


PeterDaveHello:
Please remove the strange duplicated commits first, thanks.

drewfreyling:
There are merge issues too here.

PeterDaveHello:
I think you did a wrong operation here, please open a new normal issue, thanks.

PR #4372, opened 2015-02-18
Updating ekko-lightbox package.json

ashleydw:
Modified package.json for ekko-lightbox to latest version for autoupdate but left version as 3.0.3a (current is 3.2.2) so that auto-update can do its thing.


PeterDaveHello:
Hi @ashleydw, thanks for this PR, would you please leave the indent alone and add its filename field back? So that you can pass our CI build test, thank you!

PR #4373, opened 2015-02-19
add gmaps.js version 0.4.17

threadhead:
update gmaps.js to version 0.4.17 https://github.com/hpneo/gmaps


PeterDaveHello:
Would you please also add git auto-update config for it as #3638 said?
And I think we should not touch its repo url in the same commit, thanks!

PeterDaveHello:
BTW, I noticed that there is no minified file under its repo, but it's still okay to add git auto-update for it!

drewfreyling:
Looks like the `lib` folder will be removed once https://github.com/hpneo/gmaps/issues/404 is fixed. So I'll leave this here until NPM support is added.

PeterDaveHello:
And we have conflicts here, need to rebase with our latest master branch.

drewfreyling:
Closing due to no response.

PR #4381, opened 2015-02-20
[new][author] Added webSqlTracer library

viskin:
This is a second try to add webSqlTracer library.
Origin of the file: https://github.com/terikon/webSqlTracer.git, file name: dist/webSqlTracer.min.js


PeterDaveHello:
Thanks.

PR #4390, opened 2015-02-24
[new] componentjs

AuspeXeu:
Added componentjs library


ajax/libs/componentjs/package.json
```diff
@@ -0,0 +1,91 @@
+{
+    "name":        "componentjs",
+    "npmname":     "componentjs",
+    "npmFileMap": [
+        {
+          "basePath": "/bld/",
```
PeterDaveHello: I can't find this path under its folder

AuspeXeu:
Removed that part.

PeterDaveHello:
But we still need auto-update config, is it possible? thanks!

AuspeXeu:
That should do, right? It's a different build repository I added it under the repositories property but left the repository property as is to point to the source repo.

PeterDaveHello:
Thanks, looks okay, would you please squash these commits into only 1 commit as our document said? Thank you!

PR #4392, opened 2015-02-24
Autoupdate

isaacewing:
added SVG Magic jQuery plugin


PeterDaveHello:
This is not the right way that we add a lib, would you please refer to other PRs and our document? thanks!

PR #4402, opened 2015-02-26
Update leaflet-providers to 1.0.17 and auto update.

jieter:
Update leaflet-providers to 1.0.17 and auto update.

ref leaflet-extras/leaflet-providers#132


PeterDaveHello:
Please squash these commits into only one commit as our document said, thanks!

jieter:
@PeterDaveHello done.

I thought I would get away without downloading this huge repo to my computer... No problem though.

PeterDaveHello:
There is no minified file under the repo in your auto-update config @@

PeterDaveHello:
There is a strange situation here, so I manually update the lib and add git auto-update config for it, and leave npm auto-update alone, and will debug it when I have time.

Though we have npm auto-update earlier, I don't thinks it's the old way, it's more efficient when git auto-update has no any hacks.

jieter:
Hmm, good point. Would really like to avoid adding minified versions to the project repository. Is npm update deprecated?

And why didn't it work in the first place?

PeterDaveHello:
As I said in the commit log, we got strange problem with this lib, I need to have time to debug it.
Currently, just transfer to git auto-update.

PR #4403, opened 2015-02-27
[new]knockout-postbox

davidbroadfoot:


davidbroadfoot:
Sourced from https://github.com/rniemeyer/knockout-postbox/blob/master/build/knockout-postbox.js

PeterDaveHello:
Please squash these commits into only one commit as our document said, thanks!

drewfreyling:
Also, you shouldn't have grabbed the file from master either.  Use one of the git tags instead.

PeterDaveHello:
@drewfreyling is right!

PR #4407, opened 2015-02-28
[author] jquery.scregal update auto-update file and add some files

kunass2:


drewfreyling:
If the files are live, we can't have you moving files around which will cause 404s for our users.

PR #4409, opened 2015-02-28
[author] jquery.scregal update package.json and add some files

kunass2:


drewfreyling:
@kunass2 You shouldn't be deleting stuff from the Sortable library here.

PR #4416, opened 2015-03-05
Minify foundation.min.js (5.5.1)

Vithon:
It says it's minified in the name, but it actually isn't. This just changes that.


PeterDaveHello:
Would you please use only one commit to fix this and move the original one to foundation.js ? thanks!

drewfreyling:
Looks like this was done in 17a7af2b3df80f3bc9c06f1d2c93d58468a4a35c

PR #4418, opened 2015-03-05
Minify foundation.min.js

Vithon:
*Actually* minify foundation.min.js

Non-minified foundation.js is already located in the js/foundation/ folder as foundation.js, so I didn't add it a second time.

Thanks.


PeterDaveHello:
Thanks!

PR #4419, opened 2015-03-06
[author] [new] add free-jqgrid 4.8.0

OlegKi:
add "free jqGrid" 4.8 developed in https://github.com/free-jqgrid/jqGrid
to cdnjs.


PeterDaveHello:
Thanks, it'll be better to point to the minified file next time.

OlegKi:
Thank you! I planed to include minimized versions of all JavaScript and CSS tiles in the next version, but because I generate .map files for all minimized files, then it will more increase the total size of posted files as decries it. Nevertheless I plan to do this in the next version. Thanks!

PeterDaveHello:
:+1:

PR #4420, opened 2015-03-07
Update jquery.Jcrop.css

zairwolf:


PeterDaveHello:
I don't think this changing is necessary.

PR #4429, opened 2015-03-12
[new] [author] Add savvior 0.4.2 (with autoupdate)

attila:
Please add savvior.js version is 0.4.2, the latest.
Source files are from https://github.com/attila/savvior
Added git autoupdate.


PeterDaveHello:
Thanks!

PR #4435, opened 2015-03-13
update outdated-browser to 1.1.0

CitySim:


PeterDaveHello:
We don't serve html files, and would you please tell us the source? It'll be better to add auto-update config for it.

CitySim:
source:
* https://github.com/burocratik/Outdated-Browser/tree/master
* https://github.com/burocratik/outdated-browser/archive/master.zip

is there any documentation on how to add autoupdate for non-npm modules? and i removed the html files

PeterDaveHello:
I think you should pull from v1.1.0, not the master, and would you please also add auto-update for it (#3638)? thanks!

CitySim:
source is now: https://github.com/burocratik/outdated-browser/archive/1.1.0.tar.gz
I left the old files (1.0.1) named "outdated**B**rowser" but changed the filename in the `package.json` to lower-case, not sure if this will cause problems

PeterDaveHello:
Oh, it's really nice, thanks @CitySim :+1:

PR #4436, opened 2015-03-13
[author] angular-route-segment 1.4.0

artch:
https://github.com/artch/angular-route-segment/releases/tag/v1.4.0


PeterDaveHello:
Possible to add auto-update config for it?

PR #4442, opened 2015-03-14
[author] [new] react-cookie

eXon:
Load and save cookies within your React application

The dist files are directly from the npm package. They have been built using `npm run build`.


PeterDaveHello:
:+1:

PR #4443, opened 2015-03-15
Pushing toastr to 2.1.1

TimFerrell:


drewfreyling:
https://github.com/cdnjs/cdnjs/commit/69ee0d90bf725e66cab93d7610e7caf5bb24c155 already adds 2.1.1. We use an auto update mechanism as soon as toastr tags their releases so we shouldn't need any manual submissions.

TimFerrell:
69ee0d9 does not follow the same file/folder convention as previous releases, and does not update the latest folder.

drewfreyling:
@TimFerrell ah good catch. A few things though:
* We don't serve up `scss` and `less` files so these aren't needed here - Are you able to remove them from this PR (I know they are in previous releases here but they 403 anyway)?.
* Is there a reason to split the release out into css and js folders when the `/build` folder doesn't have them split up like this in the toastr repo?
* The build is failing as `package.json` needs to be update with filename to point to the new `/js/toastr.min.js` location

Thank you for updating the latest folder as this was missed (However I'm totally against the concept of a latest folder, especially when the files are cached for a year so potentially your users could be up to a year behind the latest release - just my 2c)

PeterDaveHello:
We should use the new file/folder structure, as auto-update config do.

TimFerrell:
No issues removing the scss and less files.

I can move to the new folder structure if that's the mandate now, but given that some of our users are using the latest folder, it'd be nice if we could keep that active. 

That said, how should we move forward?

PeterDaveHello:
I think we can keep the structure in `latest` folder only, use the new structure with other newer versions.

drewfreyling:
I think that is best moving forward @PeterDaveHello 
Are you ok with this @TimFerrell ?

TimFerrell:
Yep, that's fine.

Other than the update to the `package.json` what do you need from me?

drewfreyling:
* less and scss files removed
* 2.1.1 folder to be changed to use same structure as `/build` folder on toastr.js's repo

PeterDaveHello:
I think the only thing you need to do is to copy the already exist files from 2.1.1 to latest with the origin structure. It'll be nice if you can reabse your branch with our latest master branch.

TimFerrell:
Changes should be good now - let me know if this works.

PeterDaveHello:
Not at all, changing 14 files in the PR makes no sense, I just fixed it in bbb5dff.

drewfreyling:
Thanks @PeterDaveHello 
@TimFerrell please feel free to open another PR for  088e483 to get non-minified files and add yourself as a maintainer in `package.json`.

PR #4444, opened 2015-03-16
add postscribe v1.3.2

kkshetty:
Adding the latest version of postscribe.min.js, which is v1.3.2.

Also added updated package.json


kkshetty:
Will squash the two commits into one, and then create another pull request.

Thanks,
Krishna Kishore Shetty

PR #4446, opened 2015-03-16
update postscribe to v1.3.2

kkshetty:
Added postscribe.min.js and updated package.json for postscribe v1.3.2


PeterDaveHello:
Would you please leave the current indent and unrelated parts in package.json alone? Thanks!

kkshetty:
@PeterDaveHello I did not think beautifying the package.json would cause a problem.

I'll add the original package.json(without the indents) and commit it.

kkshetty:
Will close and raise another one.

PR #4464, opened 2015-03-19
[new] Added url.js

IonicaBizau:
Added the url.js library: https://github.com/jillix/url.js


PeterDaveHello:
Thanks.

IonicaBizau:
We (@jillix) thank you! :smile:

PeterDaveHello:
np

IonicaBizau:
@PeterDaveHello I forgot to add the `autoupdate` field in package.json. Would't it be better to modify the scripts to get updates from npm? Otherwise, we would have some redundant information...

ajax/libs/urljs/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "manipulate",
+    "browser"
+  ],
+  "author": "jillix <contact@jillix.com>",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/jillix/url.js/issues"
+  },
+  "homepage": "https://github.com/jillix/url.js",
+  "npmName": "urljs",
```
PeterDaveHello: It's already set to update from npm.

IonicaBizau: What about if I don't provide that? Shouldn't it update automatically using `name` field?

PeterDaveHello: You meant you don't provide npm package? It uses npmName not name field.

IonicaBizau: I mean, why not to use the `name` field instead of `npmName`?

PeterDaveHello: We can not make sure they are the same and if npm package exist.

IonicaBizau: Ah, I see. Thanks! :+1:

Issue #2906, opened 2014-03-30
Move bower-angular-translate-*


LordSputnik:
Hi guys,

Thanks for getting the packages hosted.

One request - could you remove the "bower" prefix from the front of the package names?  In the original github repository, this is just used to indicate that the package is compatible with the bower package manager for node.js (see http://bower.io/), so it probably shouldn't be part of the package name here at cdnjs.

Thanks!

knalli:
fyi: all 8 modules of angular-translate are available now without the _bower_-prefix on cdnjs.com. Each of them autoupdates against the dedicated bower-* repositories of us.

I think we can close this here.

PeterDaveHello:
We don't want to break the running sites, so can not move but can add a new with the right name.

PR #4469, opened 2015-03-21
[author] [new] [replace] Add all angular-translate release libraries (current 2.6.1) w/ autoupdate spec

knalli:
This re-adds all angular-translate libraries including plugins targeting the bower repositories.

Despite the already existing cdnjs libs "bower-angular-translate-*" this ones:

* will follow the common naming: angular-translate, angular-translate-loader-url, and so on
* will only auto updating against the bower shadow repositories: bower-angular-translate, bower-angular-translate-loader-url, and so on
* metadata (issues, license) are referring against the main repository at github.com/angular-translate/angular-translate

I have squashed all 8 sub-libs into one commit.


knalli:
Linking #2906

PR #4472, opened 2015-03-22
bump m8tro-bootstrap to 3.3.4

idleberg:
i keep wondering why the auto-update *never* works


PeterDaveHello:
https://github.com/cdnjs/cdnjs#enabling-npm-auto-update
 > so it'll be reasonable to delay for at most 30 hours, if you think there is a lib didn't been updated, please wait at least 30 hours for the process to audit it.

PR #4477, opened 2015-03-23
Adding react-breadcrumbs

svenanders:
Tried adding this a couple of months ago, but forgot to add the initial components. I've added the two latest versions. The 0.13.0 works with React 0.13, and the 0.2.15 works with React 0.12 and below.


PeterDaveHello:
Please remove the unnecessary files.

svenanders:
Which ones?

PeterDaveHello:
I think we need these files only:
package.json
0.2.15/react-breadcrumbs.min.js
0.13.0/react-breadcrumbs.min.js

drewfreyling:
Closing due to no response and a lack of popularity with repo.

PR #4478, opened 2015-03-23
[author][new] Ractive-Require 0.1.0

XavierBoubert:
From: https://github.com/XavierBoubert/ractive-require

**With npm auto-update**


PeterDaveHello:
Thanks, it'll be better to squash into only one commit next time.

XavierBoubert:
Thans PeterDaveHello. How can I do that? I can't pull on my computer, the repo is too large (with _--depth 1_ too). I have made my commits directly on the GitHub website but I don't think we can commit more than one file at a time.

PeterDaveHello:
I think we can only do that on local.

XavierBoubert:
Ok thank you. ;-)

PR #4481, opened 2015-03-23
[new] Adding ng-slider component

jwils0n:
Downloaded from https://github.com/darul75/ng-slider


drewfreyling:
Are you able to squash these into one commit? I'm also thinking we should keep the files in the same structure as in `dist` so that it will make npm updates easier in the future.

PR #4482, opened 2015-03-23
[new] Add OpenDyslexic 2.1.0

TheTannerRyan:


drewfreyling:
@PeterDaveHello what do you think of the git auto update here? The git tags don't match the version number provided here which I'm assuming @TheTannerRyan pulled from the `bower.json` file in the repo.

TheTannerRyan:
@drewfreyling It is the correct version number seen confirmed by the original author: https://github.com/jsdelivr/jsdelivr/pull/3959

drewfreyling:
@TheTannerRyan I have to agree with @jimaek here. The author makes no mention of the current version in the readme or via tags in the repo.

PeterDaveHello:
Pulling from a non-official repo is a bad idea I think ...

PR #4483, opened 2015-03-23
[author] [new] add ractive 0.7.1, and auto-updating

Rich-Harris:
This PR adds an up-to-date version of [ractive](http://ractivejs.org), and enables auto-updating. There is already a very old version of ractive, but the folder is called `ractive.js` - I think it'd be better to switch to `ractive` since a) that's the name on npm/bower/everywhere and b) it makes more sense alongside (not currently published) plugins like `ractive-transitions-slide`. Let me know if that's unacceptable and I'll change it to `ractive.js` instead.

The source of the files is npm; they're the same files as http://cdn.ractivejs.org/latest/ractive.js et al. It includes minified versions, and sourcemaps.

Thanks!


PeterDaveHello:
Thanks @Rich-Harris

PR #4489, opened 2015-03-25
[author] Update bignumber.js

MikeMcl:
Removed outdated early v1.x.x releases.
Removed v2.x.x releases containing critical bug.
Added v2.0.5.
Added source map and unminified version subsequent to user request.
Updated package.json version number and npm auto-update config.

Unable to squash commits as using web interface.


PeterDaveHello:
Sorry that we don't remove the old versions.

MikeMcl:
Yes, I agree that you shouldn't. This was just one of a number of bad decisions I made yesterday. I'll remake the request.

PR #4499, opened 2015-03-26
[new] [author] Add cyber-font v1.0.0

TheTannerRyan:
cyber-font is a font to help serve those who are dyslexic and have visual impairments.


PeterDaveHello:
I think you should better remove version number as a folder name in the dist folder 'cause we already have tags for version, and auto-update will not work with this style.

PR #4500, opened 2015-03-27
Added newest version of hover.css with autoupdate

AndreiTelteu:
Newest version with auto-update.


PeterDaveHello:
Thanks.

AndreiTelteu:
Welcome. And thanks you guys for this great cdn :)
When will this be available on the website ?

PeterDaveHello:
within 15 mins

PeterDaveHello:
Hmmmm ... I noticed that the structure was a little bit wrong, because the bathPath is css, so we don't need to put them in css anymore, ha.

PeterDaveHello:
I think I'll just remove the bathPath so that we can keep its structure as the old version.

PR #4505, opened 2015-03-29
Updated React Router config to point to new paths

fatso83:
The paths have changed from `/dist` to `/build/global` due to building
several targets for the 0.12 and 0.13 versions. This has prevented the
builds from entering the CDNJS repo


fatso83:
One thing I wondered about was the `filename` property - should that be updated? I left it unchanged, although the file names have changed in the last versions. It used to be react-router.min.js`, but is now `ReactRouter.min.js`. Not sure how relevant this field is, as there are two files there.

PeterDaveHello:
Thanks, I'll add it when adding new version.

PR #4511, opened 2015-04-01
Crafty JS 0.6.3

icco:
Grabbed from:

 - https://github.com/craftyjs/Crafty/releases/download/0.6.3/crafty.js
 - https://github.com/craftyjs/Crafty/releases/download/0.6.3/crafty-min.js


PeterDaveHello:
Would you please squash these commits into only one commit? And according to #3638, should also add auto-update config for it, thanks.

icco:
Yeah, I was trying to avoid cloning down the whole thing since I'm not on a stable internet connection, but I'll give it a try.

icco:
Interesting, apparently I can't squash the commits, because I have a case-insensitive filesystem, so cloning this causes uncomittable changes that I can't stash or revert. Wonderful.

icco:
Also, I don't understand how to add autoupdate for sources like crafty that don't store a compiled version of their source in their repo.

PeterDaveHello:
I think they do have in dist folder :)
https://github.com/craftyjs/Crafty/tree/0.6.3/dist

icco:
Ah, I was just looking at the master branch.

Based on the documentation in that pull request you linked me, I still don't understand what the right options would be to setup this auto update.

PeterDaveHello:
Maybe you can take this commit as an example? e0f588c
Git auto-updater will pull from tags only, we also support npm auto-update.

icco:
Yeah, sorry, I don't get it. What's the base path in relation to? Is this documented anywhere?

Crafty looks like it might be an npm package, but I have no idea what npm auto-update is.

PeterDaveHello:
For git auto-update, only #3638 for the instruction, bathPath is the relative path from the root of the project, and there is a part about npm auto-update in the readme, it'll grab versions from npm package.

drewfreyling:
Closing due to no response.

PR #4512, opened 2015-04-02
Add KaTeX v0.3.0

xymostech:
Updating to the newest version of KaTeX.

[Link to the release page](https://github.com/Khan/KaTeX/releases/v0.3.0)


PeterDaveHello:
Thanks, is it possible to add auto-update for it?

xymostech:
We don't currently store built files in the repository, so not at the moment. We might start doing that at some point, and we'll update cdnjs to auto-update when we do.

PeterDaveHello:
Okay, thanks.

PR #4514, opened 2015-04-02
Autoupdate ionicons

tomsterritt:


tomsterritt:
Somewhat confused now. I was previously told updating autoupdate config would be sufficient. What's wrong?

PeterDaveHello:
I think you should take a look at the error message first, there is no version v2.0.1 in the repo, so you should not touch the version number here.

tomsterritt:
I've reverted the change in version number. To clarify then, if I don't supply a new version of the code as well I don't update the version number, _only_ the autoupdate section?

PeterDaveHello:
Yeah, the version field will be updated once we update with the real files.

PeterDaveHello:
I think you should keep the origin structure in cdnjs, there are 4 folders:
css/  fonts/  png/  src/

moesalih:
What's the status on this?

PeterDaveHello:
Manually picked in 9319fe8 Autoupdate ionicons Tom Sterritt 4 months ago

PR #4521, opened 2015-04-03
[new] Added jquery-form-serializer@1.1.0

IonicaBizau:
We use this library internally at @jillix and open-sourced it recently: https://github.com/jillix/jQuery-form-serializer


PeterDaveHello:
Please also add minified file as our document said, thanks!

IonicaBizau:
@PeterDaveHello Sorry, I just added it. Would be nice to automatically generate them if they are not minified.

Thanks! :+1:

PeterDaveHello:
Yeah, but is not very easy to determinate if is need, post-fix .min.js or .min.css are not used by every libs, they also use .src .dev -min .prod ... blah blah blah, even the filename without that post-fix may be also minified lol ...

IonicaBizau:
That's also true. :+1:

PR #4523, opened 2015-04-03
[author][new] Added validate.js@0.6.1

ansman:
The files were downloaded from https://github.com/ansman/validate.js


PeterDaveHello:
Thanks.

PR #4524, opened 2015-04-03
[author] [new] Added geojs@0.4.1

jbeezley:
New library from [github](https://github.com/OpenGeoscience/geojs) and [npm](https://www.npmjs.com/package/geojs).  The main file committed here is from https://github.com/OpenGeoscience/geojs/blob/v0.4.1/geo.min.js.


PeterDaveHello:
Thanks.

PR #4535, opened 2015-04-07
[new] CSS Spinning Spinners v1.1.0

BillyNate:


BillyNate:
Contains eight equally important files, what should the filename mentioned in the package.json be?

PeterDaveHello:
Missing field - filename, which will point to the main file.

BillyNate:
I understood the missing field, that's why I said that all files are equally important. There is no main file.

So, just should I just set the first file as the main file to make it pass the test?

PeterDaveHello:
Yeah, please.

BillyNate:
Filename property added, check passed and commits squashed.

PR #4545, opened 2015-04-10
Added Authy Form Helpers v2.3

senekis:
from https://github.com/authy/authy-form-helpers/tree/v2.3


PeterDaveHello:
Would you please also add auto-update config for it? So that we can easily pick other versions.

PR #4563, opened 2015-04-14
Added jquery.scroll4ever using the new way

luanlmd:
There we go.


PeterDaveHello:
Would you please also add its real files as the first commit to add a lib? So that you can pass the test.

luanlmd:
Do I need to add the files just the first time if I'm using the autoupdate method? And should I redo the initial commit or sync the upstream repo and do another commit to add the files?

PeterDaveHello:
Just the first time, you can do it in the same commit, and use force push to update this PR, thanks!

luanlmd:
Looks like it is all right now.

PeterDaveHello:
:+1:

PR #4576, opened 2015-04-23
Make repositories field in package.json be consistently a list

yuvipanda:
Most other package.json's seem to have repositories be a list with dicts in them,
and the name being 'repositories' also makes sense with it being a list.


yuvipanda:
The commit messages probably need to be fixed. I can also squash this up if you folks want.

PeterDaveHello:
It'll be good if you can make them in the same format, please.

yuvipanda:
@PeterDaveHello I squashed them and made a single commit message

PR #4577, opened 2015-04-24
Add terra.js

rileyjshaw:
Source: https://github.com/rileyjshaw/terra.git


PeterDaveHello:
Don't you wanna add it now? @rileyjshaw

rileyjshaw:
@PeterDaveHello I actually realized that it's hosted elsewhere, and since it's not on npm auto-updating would be a bit tricky. Overall I don't think that it's a great fit.

Thanks for checking, though!

PeterDaveHello:
We also support git auto-update, if you want.

rileyjshaw:
@PeterDaveHello that's alright - thanks though!

In the meantime, I think [this](https://github.com/cdnjs/cdnjs/pull/4578) one's ready to go

PR #4587, opened 2015-04-27
Added lz-string

pieroxy:


PeterDaveHello:
Please squash them into only one commit and tell me the source of the files as our document said, thanks.

pieroxy:
For point 1, I'm sorry I don't have enough resources (in time, bandwidth and computational power) to clone a 11GB repo locally, squash my PR and push it back. Any git command takes more than 4 minutes before I lose patience and Crtl-C it. What I can do is re-clone the repo and push all 4 files independently (that is, squash the package.json commits). Would that do?

As for the source of the files, I'm the author and creator of the lib. What else do you need? (For the record, your document don't say anything mandatory as to what has to be included in a PR...)

Thanks for letting me know if a PR with 4 commits for the 4 files is acceptable.

pieroxy:
Done. Took a while. Unable to do it on Windows or on a non-SSD machine. But worked in the end, from work.

PeterDaveHello:
Thanks, would you please provide minified base64-string.js as base64-string.min.js next time?

PR #4588, opened 2015-04-27
Added knockout-pre-rendered@0.3.0 [author] [new]

ErikSchierboom:


ErikSchierboom:
The build fails, but looking at the log does not make me any wiser as to it is problem of my PR or a general problem. I do see a lot of other failed builds, so maybe it is something general?

PeterDaveHello:
Thanks, it'll be better to set filename field to `knockout-pre-rendered.min.js` next time.

PR #4589, opened 2015-04-27
Remove reference to cdnjs.cloudflare.com

ram-nadella:
The new link is now cdnjs.com and links to the https site.


ram-nadella:
Never mind, I just realized there are a few more references to this. Will update pull request after I catch the others.

PR #4592, opened 2015-04-28
[author] rename jquery-sheetrock to sheetrock and add 1.0.0

chriszarate:
Not sure what the preferred method of doing this is. As of `1.0.0`, my package no longer depends on jQuery so has been renamed from "jquery-sheetrock" to "sheetrock." The file names also changed to remove the "jquery" identifier. Further, the underlying NPM package has been renamed as well—or, more accurately, a new NPM package named "sheetrock" was created and the existing "jquery-sheetrock" package was deprecated.

I'm pretty sure all this broke/confused the auto-updater, so I'm updating the `package.json` and adding files for `1.0.0`.


chriszarate:
Ok, looks like I can't change the package name in package.json. Closing and will resubmit.

chriszarate:
Thanks!

PR #4596, opened 2015-04-29
[new] Adding page.js

jcoffland:
I'm trying to add [page.js](https://github.com/visionmedia/page.js) as an auto updated library.

Did I do this right?  What about minifying and the library dependencies?  Are these handled automatically?  The documentation is inconsistent.


jcoffland:
This didn't work.

PR #4597, opened 2015-04-29
[new] Adding page.js

jcoffland:
Trying this again with autoupdate.


jcoffland:
The tests failed but shouldn't it automatically pull in the latest version?

PeterDaveHello:
The first PR to add a lib should contain is real files.

jcoffland:
Ok, thanks, I'll try that.

jcoffland:
The original Javascript in git is incomplete.  It has dependencies on other packages and is not minified.  I don't have control of the original project.  Are there ways to solve this?  Is it possible to automatically pull from npm archives?

PeterDaveHello:
We support npm auto-update.

jcoffland:
@PeterDaveHello Any idea why this is still failing?

PeterDaveHello:
As it said ...

>       ✗ page.js package.json is valid 
        » page.js didn't parse as any known format: 
  	  » NPM package.json 
  		Property is required (true): /properties/filename // valid-packages-test.js:82

jcoffland:
Ok, I saw that but didn't understand it.  After some thought I think I get what it means.

PeterDaveHello:
And please squash these commits into only one commit and rebase(sync) your branch with our latest master branch, thanks.

jcoffland:
How's it look now?

PeterDaveHello:
Looks okay, it'll be even better if you can clean up the commit message, ha.

jcoffland:
Ok, I'll do that next time.  BTW, squashing is a pain.  I had done everything on GitHub up to that point.  Then I had to clone the entire repo which is current over 1GiB.  This takes awhile.  I wish there was a better way.  I suppose some friction keeps the lazy devs away.  Thanks for the merge!

PeterDaveHello:
Yeah, we need time to setup a whole new method to handle this issue. BTW, the size of repo should be only about 800MB.

jcoffland:
If I run ```du``` on my clone it says it's using 12GiB.  When I cloned the repo it downloaded 624.96 MiB in objects according to git.  So you're right the download is less than 1GiB but still large.

Yeah, I'm sure a new method would be a lot of work.  Thanks for creating this project, BTW.

PeterDaveHello:
np

PeterDaveHello:
Oh BTW, I think you can add `--depth 10` to decrease the repo size because you don't need the history and old, dirty things that we already dropped.

jcoffland:
I read that doing this can make your push fail.  Haven't tried it though.

PeterDaveHello:
I don't think so, I use it many times.

jcoffland:
Good to know.

PeterDaveHello:
BTW, I think this will help you:
https://github.com/cdnjs/cdnjs/blob/master/sparseCheckout.md

jcoffland:
Thanks!

PR #4607, opened 2015-05-03
add new library

mogafk:
add new library


drewfreyling:
@mogafk This library isn't popular enough for us to add to the CDN.

PR #4612, opened 2015-05-04
update URI.js to use auto update

codylerum:


PeterDaveHello:
What about other js files?

codylerum:
@rodneyrehm which files should be published?

rodneyrehm:
does [package.json's "files"](https://github.com/medialize/URI.js/blob/gh-pages/package.json#L49-L61) answer that?

codylerum:
@rodneyrehm Yes it does. I'll update the PR. Thanks!

codylerum:
updated.

PR #4624, opened 2015-05-09
[new] [author] Dragdealer@0.9.8

skidding:


PeterDaveHello:
Hi @skidding, would you please squash these commits into only one commit as our document said? thanks!

skidding:
@PeterDaveHello I would if I could...

The repo is so big I can't do any git operation on it. It takes up 11GB and it took >30 minutes to clone. `git status` takes ~30 seconds and I can't get to rebasing because I always get an untracked file no matter how many times I reset/checkout. I'm using an old i5 MacBook Air, but still...

The reason why I had so many commits is because I used the GitHub UI where you can only add one file at a time (commit).

Tips?

PeterDaveHello:
Sparse checkout will help:
https://github.com/cdnjs/cdnjs/blob/master/sparseCheckout.md

skidding:
Thank you @PeterDaveHello for the support. All good?

PeterDaveHello:
It'll be good if you can also setup auto-update config for it!

PR #4635, opened 2015-05-13
[new] Added ngreact@0.1.5

eseceve:


drewfreyling:
I'm not sure we serve up `.jsx` files here. Tagging @PeterDaveHello for clarification.
Can you also squash your commits please?
Also, you will need to have a look at your npm map, adding .* will include files that we don't serve as well.

PeterDaveHello:
I am not sure, and what's the source of the files?

drewfreyling:
@thomasdavis do you know if we support .jsx files? @eseceve where did you get these files from?

drewfreyling:
closed due to no response.

PR #4637, opened 2015-05-14
Enable autoupdate in sockjs-client package.json

brycekahle:


PeterDaveHello:
I think the change of maintainer should be in another commit, what do you think?

brycekahle:
I think the one line is not worth another commit. Especially because it is a metadata file.

PeterDaveHello:
Yeah, but the commit message is `Enable autoupdate in sockjs-client package.json`, is not about worth or not, the change didn't match the commit message, I think you can also add a line in the commit message to talk about the change.

brycekahle:
Updated to 2 commits as requested

PeterDaveHello:
Thanks!

PR #4652, opened 2015-05-19
[new] Add Layzr.js

irazasyed:
Origin: https://github.com/callmecavs/layzr.js

I used the **cdnjs-importer** and then added the auto-update config manually. Wouldn't it be good if the importer automatically adds the auto-update config too?


PeterDaveHello:
Yeah I think we should ask the author @IonicaBizau.

ajax/libs/layzr.js/package.json
```diff
@@ -0,0 +1,50 @@
...
+      "files": [
+        "**/*"
+      ]
+    }
+  ],
+  "filename": "layzr.js",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/callmecavs/layzr.js.git",
+    "basePath": "",
```
PeterDaveHello: should be `dist`

PeterDaveHello:
I noticed that you add two auto-update config for it, please remove one of them, maybe remove the npm config is a good idea and leave git auto-update alone, btw, filename should point to the minified file, thanks.

IonicaBizau:
@irazasyed @PeterDaveHello Hey! The *cdnjs-importer*  tool already is supposed to do this. It doesn't use `autoupdate` but `npmFileMap` (which is already there).

Maybe I misunderstood the things, but isn't `npmFileMap`used in the auto update scripts? [Here](https://github.com/cdnjs/cdnjs/blob/cfee0868eebc8c1418444fae8f5a410f8b5a9290/ajax/libs/svg.connectable.js/package.json#L11-L18) is an example. I saw that `svg.connectable.js@1.1.0` was added automatically. :sparkle: So, should *cdnjs-importer* add `autoupdate` as well?

PeterDaveHello:
yes, npmFileMap is used for auto-update, so seems we don't have issue here, that's the reason why there is 2 auto-update configs here ... because one of them was added by **cdnjs-importer**, cool!

irazasyed:
Well, I didn't notice that. So i just need to remove the git auto-update right? @PeterDaveHello

IonicaBizau:
@irazasyed Exactly! :sparkles:

PeterDaveHello:
yep

irazasyed:
@PeterDaveHello @IonicaBizau Alright, Done. Thanks for confirming!

PeterDaveHello:
hmmmm ... should use `rebase` to remove the commit in this case, please.

irazasyed:
Yep, I'm on it and trying. Git freezes and takes so much time to respond for some reason (Very weird).

PeterDaveHello:
You can try `git reset HEAD^` for two times to revert it and then force push.

irazasyed:
@PeterDaveHello Alright, That worked. Thanks :+1:

PR #4656, opened 2015-05-21
PNotify 2.0.1, autoupdate

chrillep:


PeterDaveHello:
Should not bump version here, and you point the filename to a file not on the auto-update file map, is that correct? What's the reason why to change the filename field here?

chrillep:
ok, misinterpreting the docs.
pnotify.custom.min.js is what the creator uses when they are minified and combined 
pnotify.core.js is just one part ...

skarmavbild 2015-05-21 kl 15 03 46

PeterDaveHello:
then you should not refer to a non-exist file.

Piicksarn:
@cdnjs/intern @cdnjs/intern2

maruilian11:
@chrillep please squash commits to one clear commit, thanks!
the PR will be close if there is no response in next 7 days.

PeterDaveHello:
Auto-update setup in #4933 https://github.com/cdnjs/cdnjs/commit/c0a48209971af2aa01d3261fdefb5ff2860d39fb

PR #4657, opened 2015-05-21
[author] Change configuration of perfect-scrollbar.

noraesae:
Firstly, the plugin is no longer just a jQuery plugin. It works with vanilla JS and it's rather recommended. So I thought it would be better to change the name from jquery.perfect-scrollbar to perfect-scrollbar.

Also, the main repo no longer provides the built files. It's provided by its bower repo, so the auto update configuration of it has been changed too.


PeterDaveHello:
Sorry that we can not 'change' the name of it, it'll brake the current users using this lib, maybe you can create a new folder for it?

noraesae:
Oh, surely I can. Will update in a while.

noraesae:
Or I may just stick with the original name. I think it's not a big deal to keep the `jquery.` prefix. Anyway, will update soon!

PeterDaveHello:
It's up to you I think.

noraesae:
Updated.

PR #4671, opened 2015-05-24
[author] Update ifvisible version

serkanyersen:
Ifvisible.js is on npm now, however name on npm is different from cdnjs

npm => `ifvisible.js`
cdnjs => `ifvisible`

Will auto update still work? should I include something in the package.json?

Thanks


PeterDaveHello:
Should add auto-update config for it @serkanyersen

PeterDaveHello:
And should better squash them into only one commit next time, thanks.

PR #4672, opened 2015-05-24
[author] ifvisible - adds npm auto update configuration

serkanyersen:
Adds npm configuration

Thanks

### DIFF
--- a/ajax/libs/ifvisible/package.json
+++ b/ajax/libs/ifvisible/package.json
@@ -8,7 +8,7 @@
   "npmFileMap": [{
     "basePath": "/src/",
     "files": [
-      "*.js"
+      "*.*"
     ]
   }],
   "keywords": [
--- a/ajax/libs/ifvisible/package.json
+++ b/ajax/libs/ifvisible/package.json
@@ -4,6 +4,13 @@
   "version": "1.0.3",
   "description": "Crossbrowser & lightweight way to check if user is looking at the page or interacting with it. (wrapper around HTML5 visibility api)",
   "main": "ifvisible.js",
+  "npmName": "ifvisible.js",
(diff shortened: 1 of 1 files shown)


ajax/libs/ifvisible/package.json
```diff
@@ -4,6 +4,13 @@
   "version": "1.0.3",
   "description": "Crossbrowser & lightweight way to check if user is looking at the page or interacting with it. (wrapper around HTML5 visibility api)",
   "main": "ifvisible.js",
+  "npmName": "ifvisible.js",
+  "npmFileMap": [{
+    "basePath": "/src/",
+    "files": [
+      "*.js"
+    ]
```
PeterDaveHello: What about the map file?

serkanyersen: Previous versions didn't have the map file so I didn't add it here.

telkomsel13:
#1

ajax/libs/ifvisible/package.json
PeterDaveHello: But it has now, don't you add it?

serkanyersen:
@PeterDaveHello Updated now to include all files under `src/` map file is for coffee file so it needs to be there as well.

Thanks

PeterDaveHello:
@serkanyersen: we don't support coffee script on cdn, so should not include them.

Piicksarn:
@cdnjs/intern @cdnjs/intern2

maruilian11:
@serkanyersen any update? thanks!

PeterDaveHello:
Auto-update setup in 120e783 #6564.
@maruilian11 @Piicksarn you should check what do we need to here first, instead of meaning less comment.

PR #4675, opened 2015-05-24
OWNER: Update textAngular File structure.

SimeonC:
All our production files are only in the dist folder now.

### DIFF
--- a/ajax/libs/textAngular/package.json
+++ b/ajax/libs/textAngular/package.json
@@ -26,7 +26,6 @@
     "target": "git://github.com/fraywing/textAngular.git",
     "basePath": "",
     "files": [
-      "src/*",
       "dist/*"
     ]
   }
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
What about also change `basePath` to dist? So that the file map can be `**/*`.

SimeonC:
I could, but it has no subfolders so do I need to? Is that more inline with the other packages?

PeterDaveHello:
There is no two folders there, so I think we don't need to use empty bashPath and make the structure so long, bashPath will be the root of each version, `**/*` instead of `*` just help us no need to update it again once it has sub folders.

SimeonC:
Sounds good, done that now.

PeterDaveHello:
Ping @SimeonC, any comments?

SimeonC:
Can't remember. I think this was a major change anyway that people constantly got confused about so I think this is fine as is.

PeterDaveHello:
Okay, got it.

PR #4679, opened 2015-05-27
Adding Watermarkjs [new]

rubenpa:
Watermarkjs github page is here: https://github.com/brianium/watermarkjs
Project page: http://brianium.github.io/watermarkjs/


drewfreyling:
Your package.json isn't valid - it is missing `filename` - which is why the build isn't working.

PR #4682, opened 2015-05-28
[author] HIW JavaScript API v5.0.2, v5.0.3, and v5.0.4.

JoshMcCullough:
http://developers.healthindicators.gov


PeterDaveHello:
What's the reason why to remove auto-update?

JoshMcCullough:
I wanted to clarify how it works. E.g. I assume it should point to a "current release" branch which only ever gets updated when a release is done. But how does cdnjs know the version to use when it detects changes in the branch?

PeterDaveHello:
It'll grab files from each git tags as its versions.

JoshMcCullough:
So does it just monitor the repo for commits and if it finds commits, looks for tags which look like a version number?

JoshMcCullough:
I'm going to update this PR with auto-update reinstated.

PeterDaveHello:
It looks into all the tags.

JoshMcCullough:
PR updated with autoupdate.

PeterDaveHello:
Should set `basePath` to `Content/Scripts/API/Output/` instead of adding it to all the filename map

JoshMcCullough:
Should be good to go now. Sorry for all the back-n-forth.

PR #4699, opened 2015-06-01
Added alasql lib

mathiasrw:


PeterDaveHello:
Don't we need other files under dist?

mathiasrw:
For the moment the abillity to include "plugins" from the other files is considered an experimental feature that is not documented yet - so nope. Only the alasql.js is needed (and the worker version if one wants to use it as a webworker)

mathiasrw:
(but thanks for keeping an eye out)

mathiasrw:
:+1:

PR #4708, opened 2015-06-04
jquery.jscrollpane.min.js v2.0.22

pavelsl:
New version of jquery.jscrollpane.min.js (2015-04-25)


PeterDaveHello:
Please read the document for what you need to send a PR.

Issue #4627, opened 2015-05-09
[author] Question about the procedure for renamed libraries


FagnerMartinsBrack:
Hi, we recently renamed [jquery-cookie](https://github.com/carhartl/jquery-cookie) to [js-cookie](https://github.com/js-cookie/js-cookie) to remove jquery dependency. We decided not to physically rename the repository and 2 new versions were added in js-cookie releases as a continuation of jquery-cookie development (1.5.0 and 1.5.1, check [here](https://github.com/js-cookie/js-cookie)). To follow semver spec we did added new features for 1.5.0 but maintained backwards compatibility with 1.4.1 (that compatibility will be removed in [v2.0.0](https://github.com/js-cookie/js-cookie/releases/tag/v2.0.0-beta.1) on js-cookie namespace).

Is there any way to setup an automatic redirect to the js-cookie repository without having to create a new one? What would be the correct procedure here?

Thanks.

PeterDaveHello:
Currently, we don't have the procedure to handle this issue, cc @thomasdavis & @ryankirkman.

FagnerMartinsBrack:
@PeterDaveHello Any news on this one?

I believe it's worth creating a new library and then setting up a message in jquery-cookie page saying "this project was moved to js-cookie" or something like that.

What you guys think?

PeterDaveHello:
hmmmmm ... I have no idea ...

FagnerMartinsBrack:
Ok, but is it viable to implement something that might allow us to provide a message in jquery-cookie CDN page? We will add a new project.

ryankirkman:
We can put a message in the package description so it shows up on the
library page.

PeterDaveHello:
Symbolic link will work https://github.com/cdnjs/cdnjs/issues/5287

MattIPv4:
No further progress, closing.

Going forward, I think we'll achieve this normally by simply adding renamed libraries as if they were a new library.

PR #4701, opened 2015-06-01
added git autoupdate to processing.js

pzp1997:

PR #4713, opened 2015-06-04
[new] [author] Add js-cookie (former jquery-cookie)

FagnerMartinsBrack:
jQuery dependency was made optional in the version `1.5.0`. `js-cookie` repository contains [all `jquery-cookie` releases/tags](https://github.com/js-cookie/js-cookie/releases) since v1.

X-Ref: https://github.com/cdnjs/cdnjs/issues/4627


FagnerMartinsBrack:
> Property is required (true): /properties/filename

It requires the "filename" property right? I see people [use](https://github.com/cdnjs/cdnjs/pull/4701/files#diff-56e8021a2281e45c7dac5f358e79d09eL3) the [filename](https://github.com/cdnjs/cdnjs/pull/4704/files#diff-3e3b074fc16b67765297884fb46029efL5) to specify the minified resource, but js-cookie, as a source repo, doesn't expose the distribution file. In this case, what should I put in that field?

> For those libs can use auto-update, you should add auto-update config for it, but as the first pull request to add a lib, you should still add its real files, or you won't pass the test.

...

> ./ajax/libs/js-cookie/2.0.0-pre/undefined does not exist but is referenced in package.json

Do I have to add all tags minified files? :tired_face:

PeterDaveHello:
I think you should put the real files as the first PR to add a lib.

FagnerMartinsBrack:
Is it ok to add the unminified files?

PeterDaveHello:
It's fine.

FagnerMartinsBrack:
@PeterDaveHello 

I made your work easier and added the minified files directly from the build of each version tag.

I have added only the versions `>= 1.5.0`. Those are the versions available only in the js-cookie project, it didn't make any sense to add the versions that already exist under jquery-cookie name.

It would be very nice to add a notice in [jquery-cookie](cdnjs.com/libraries/jquery-cookie) page on cdnjs to highlight the importance to use the latest supported version under the name js-cookie.

PR #4720, opened 2015-06-09
[new] Added later@1.1.6

MMore:
See https://github.com/bunkat/later

PR #4721, opened 2015-06-09
[new] Added prettycron@0.10.0

MMore:
See https://github.com/azza-bazoo/prettycron. `prettycron.js` minified using Uglifier to `prettycron.min.js`.


PeterDaveHello:
Same problem as #4720 ...

PR #4722, opened 2015-06-09
[new] [author] Added hprose-html5@1.5.3

andot:
Adding hprose-html5.js from https://github.com/hprose/hprose-html5


ajax/libs/hprose-html5/package.json
```diff
@@ -0,0 +1,61 @@
...
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/hprose/hprose-html5.git",
+    "basePath": "dist/",
+    "files": [
+      "hprose-html5.js"
+    ]
+  },
+  "npmName": "hprose-html5",
```
PeterDaveHello: No need to have npmName and npmFileMap 'cause you have git auto-update config.

andot: I didn't write that code, it is added by cdnjs-importer. Need I to remove it?

PeterDaveHello: hmmmm ... yes, should remove it.
cc @IonicaBizau

IonicaBizau: Who added the `autoupdate` field? The importer tool only adds `npmName` and `npmFileMap`.

PeterDaveHello: hmmmmm .... funny, but i prefer to use git auto-update, it's more reliable in fact, I still don't have time to rewrite npm auto-update.

IonicaBizau: OK, so actually the importer should be even smarter: to check if the `autoupdate` field is there or not. If it is, then it will **not** add the `npmFileMap` in the `package.json` file, right?

PeterDaveHello: Actually, I don't think there will be autoupdate field by default, maybe we can make the importer support importing from git repo?

IonicaBizau: I don't understand what you mean. The importer already gets the `package.json` from repo and reads it.

PeterDaveHello: just like it grabs files from npm, just grab from git repo

IonicaBizau: @PeterDaveHello Open an issue on cdnjs-importer so we can discuss there. Thanks!

PR #4729, opened 2015-06-11
[author] Add WinJS 4.0.0 and 4.0.1

rigdern:
These bits came from:
  - 4.0.1: https://github.com/winjs/winjs/releases/download/v4.0.1/winjs.zip
  - 4.0.0: https://github.com/winjs/winjs/releases/download/v4.0.0/winjs.zip


rigdern:
Is autoupdating the preferred mechanism yet? The readme made it sound like it was but I looked at a couple of popular projects and they didn't seem to have autoupdate enabled.

If autoupdating is preferred, I'd be interested in seeing some instructions for how to set it up.

PeterDaveHello:
Should setup auto-update.

rigdern:
@PeterDaveHello I added another commit which sets up git autoupdate.

PeterDaveHello:
Please use git protocol instead of ssh, and you did something unrelated with auto-update config in that commit, will be better to remove them, thanks.

rigdern:
@PeterDaveHello I pushed a new version. Here's what I changed:
  - Split changes to package.json into 2 commits (one for updating to 4.0.1 and one for the auto update changes)
  - Switched to the git protocol instead of ssh

PR #4745, opened 2015-06-16
[new] added inheritance.js

bsara:
Inheritance.js: https://github.com/bsara/inheritance.js


ajax/libs/inheritance.js/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "inheritance.js",
+  "npmName": "inheritance-js",
+  "npmFileMap": [
+    {
+      "basePath": "/dist",
+      "files": [
+        "*.js",
+        "src/*.js"
```
PeterDaveHello: You didn't have `src` in the PR.

bsara:
Does it not flatten the file hierarchy? I thought it did. If not, then I can make some changes.

ajax/libs/inheritance.js/package.json
bsara: @PeterDaveHello I thought that the file hierarchy got flattened upon update. Is that not true? If not, then I can make some changes to the library so that the files will all appear in the same directory.

PeterDaveHello: No, the basePath will be the root, so it will be ignored, src will not.

PeterDaveHello:
And please also squash them into only one commits, thanks!

PR #4747, opened 2015-06-17
Update project algoliasearch to 3.6.1

LeaYeh:


PeterDaveHello:
Since there is already npm auto-update config, manual update will not be accepted, we'll use auto-updater to grab the new versions and commit them after auditing.
https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/algoliasearch/package.json#L27
```
  "npmName": "algoliasearch",
  "npmFileMap": [
    {
      "basePath": "/dist/",
      "files": [
        "*.js"
      ]
    }
  ]
```

PR #4757, opened 2015-06-19
add git autoupdate config

maruilian11:


PeterDaveHello:
You should talk about which lib you are working on in the commit message.

PR #4758, opened 2015-06-20
Added plottable.js@1.0.0

punya:


punya:
Sorry, I pushed this too early - going to clean it up first.

PR #4767, opened 2015-06-21
add classie to v1.0.1 with git autoupadte config

LeaYeh:


PeterDaveHello:
Typo, and conflict/duplicated with #4766, so close now.

PR #4768, opened 2015-06-21
add jquery.actual 1.0.16 with git autoupdate config

LeaYeh:


PeterDaveHello:
I don't think we need package.js, do we?

PR #4773, opened 2015-06-23
add vue-resource v0.1.3 with git autoupdate

maruilian11:


ajax/libs/vue-resource/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "autoupdate": {
+  "source": "git",
+  "target": "git://github.com/vuejs/vue-resource.git",
+  "basePath": "dist",
+  "files": [
+  "*"
+  ]
+  }
+}
```
PeterDaveHello: Wrong indent here.

PR #4760, opened 2015-06-20
Added plottable.js@1.0.0

punya:

PR #4777, opened 2015-06-23
Test autoupdate block for validity when present

punya:
This would automate some of the review feedback from #4760.


PeterDaveHello:
Nice work, thanks @punya

PR #4786, opened 2015-06-25
add t3js v1.4.1 with git autoupdate

maruilian11:


ajax/libs/t3js/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "component"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/box/t3js.git",
+    "basePath": "dist",
+    "files": [
+      "t3.min.js",
+      "t3.js"
+    ]
```
PeterDaveHello: Should at least also add `t3.min.js.map`.
I think we also need `t3-native.js`, `t3-jquery` related files.

maruilian11:
problems are fixed, please take a look, thx

ajax/libs/t3js/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "client-side",
+    "modular",
+    "component"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/box/t3js.git",
+    "basePath": "dist",
+    "files": [
+      "*"
```
PeterDaveHello: Should not use wildcard in this case.
https://github.com/cdnjs/cdnjs#conventions
> Filenames should not include a version number and be lowercase.
> This is OK: useful.min.js, but this is not: useful-2.0.1.min.js.

maruilian11:
fixed

PeterDaveHello:
t3.min.js has src, but t3-native.min.js and t3-jquery.min.js don't, let's include them to make it consistent.

maruilian11:
fixed again.

PeterDaveHello:
same same

PeterDaveHello:
any updates?

maruilian11:
yes, plz take a look again.

PeterDaveHello:
Commit's parent was too old, please rebase on our latest master branch.

maruilian11:
done.

PR #4788, opened 2015-06-25
Add glyphicons fonts.

yohankoehler:
*add fonts directory with glyphicons


PeterDaveHello:
Is it possible to apply it to auto-update config in https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/bootswatch/package.json?

IonicaBizau:
@PeterDaveHello There is the `npmFileMap`, probably added by the importer. Isn't that enough?

PeterDaveHello:
I meant not manually add these files.

PeterDaveHello:
`fonts` already exist in the config, but we can not always manually add it, so should find a method to handle this.

alanmsm:
Still return 404 at 3.3.5 version..

PR #4791, opened 2015-06-26
add more detail about bootstrap-slider's package.json

LeaYeh:


PeterDaveHello:
hmmmmm ... actually we don't need so **more**, let's add the info meaningful in our repo!

LeaYeh:
fixed it.

PeterDaveHello:
Would you please remove `scripts` and fill in `author` field? `readmeFilename` should also be removed, thanks.

LeaYeh:
check

PeterDaveHello:
You should not remove `npmName` field as our npm auto-update needs it.

LeaYeh:
done.

PeterDaveHello:
Seems you remove the old one and add a new one, what about just not to touch it? And also `homepage` field. Thanks.

LeaYeh:
check it again ><"

PR #4805, opened 2015-06-29
add Clamp.js@0.5.1 with git autoupdate

maruilian11:


PeterDaveHello:
There is no any tags in https://github.com/josephschmitt/Clamp.js, would you please also open an issue in its repo, if the author can add tags for each release, git auto-update can work, and vice versa. I'll merge this first.

PR #4824, opened 2015-07-04
add git autoupdate config in pie.

Piicksarn:
Add git autoupdate config in pie.

### DIFF
--- a/ajax/libs/pie/package.json
+++ b/ajax/libs/pie/package.json
@@ -23,5 +23,14 @@
       "type": "git",
       "url": "https://github.com/lojjic/PIE"
     }
-  ]
+  ],
+  "autoupdate": {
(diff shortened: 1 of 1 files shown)


ajax/libs/pie/package.json
```diff
@@ -23,5 +23,14 @@
...
     }
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/lojjic/PIE.git",
+    "basePath": "",
+    "files": [
+      "PIE.js",
+      "PIE.min.js"
```
PeterDaveHello: I didn't see these files under its repo's root.

PeterDaveHello:
I don't think we can apply a working auto-update on this lib, close now.

PR #4830, opened 2015-07-05
Add git autoupdate config in psd.js

Piicksarn:


ajax/libs/psd.js/package.json
```diff
@@ -22,5 +22,15 @@
...
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/meltingice/psd.js.git",
+    "basePath": "dist",
+    "files": [
+      "psd.js",
+      "psd.min.js",
+      "psd.js.map"
```
PeterDaveHello: You can set to `*` next time :+1:

PeterDaveHello:
Seems there is no min and map file in tags, would you like to open an issue ticket in its repo so that the author may put them into it in next release? Thanks.

Piicksarn:
[ There is no min and map file in tags. #33 ](https://github.com/meltingice/psd.js/issues/33)

PR #4833, opened 2015-07-06
Auto update cropper

fengyuanchen:


PeterDaveHello:
Should only touch `autoupdate` in the commit.

PR #4836, opened 2015-07-06
[new]add KindEditor v4.1.10

huaiping:
KindEditor is a lightweight, Open Source, cross browser, web based WYSIWYG HTML editor. KindEditor has the ability to convert standard textareas to rich text editing.

Downloaded from its official website http://kindeditor.org/


PeterDaveHello:
We don't serve dll, asp .. you should put supported static files only.

PeterDaveHello:
Another problem, commit's parent was too old.

PR #4839, opened 2015-07-06
Add git autoupdate config in Placeholders.js

LeaYeh:


drewfreyling:
@PeterDaveHello should `basePath` be "/dist" like it is here or just "dist"?

PeterDaveHello:
Both fine, but I think the filemap should be `*`.

LeaYeh:
check it

PR #4840, opened 2015-07-06
Add git autoupdate config in r2d3

LeaYeh:


drewfreyling:
Please specify the 2 js files rather than *.js. In case they add unwanted files in the future in the root since they don't have a dist or build directory.

PeterDaveHello:
Or `r2d3*.js` will be fine, should not use `*` under non-dist folder.

LeaYeh:
check it

PR #4844, opened 2015-07-07
Adding bootstrap-material-design 0.3.0 fonts.

westy92:
I got the added files from: https://github.com/FezVrasta/bootstrap-material-design/tree/master/dist/fonts

I also updated the package.json file to include *.woff2 files.


PeterDaveHello:
Passed test locally. CircleCI has problem now.

PR #4846, opened 2015-07-07
merged new files

soundar24:
roundSlider folder added


soundar24:
wrongly updated pull request without adding files

PR #4855, opened 2015-07-07
Add git autoupdate config in sketch.js

Piicksarn:


PeterDaveHello:
Should fix the error first.

PR #4861, opened 2015-07-08
add git autoupdate config in bucky

LeaYeh:


PeterDaveHello:
Would you please also add non-minified file? I think this lib is popular enough, and adding the src file won't break current structure, thanks.

LeaYeh:
OK

PR #4875, opened 2015-07-10
[new] Add startbootstrap-agency@1.0.4

Amomo:


ajax/libs/startbootstrap-agency/package.json
```diff
@@ -0,0 +1,29 @@
...
+      "url": "https://github.com/IronSummitMedia/startbootstrap-agency"
+    }
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/IronSummitMedia/startbootstrap-agency.git",
+    "basePath": "",
+    "files": [
+      "js/agency.js",
+      "js/contact_me.js",
```
PeterDaveHello: What is `contact_me.js` for?

Amomo: In this theme, it contains `contact_me.php` to create email and send the message.
Anyone who use this theme might also use `contact_me.js`, so I add it.

Amomo: Oh, I find `contact_me.js` is also exist in other theme!
It should be used with `jqBootstrapValidation.js` together.
I will revise the commit!

PR #4903, opened 2015-07-10
Add git autoupdate config in Pleasejs

Piicksarn:


PeterDaveHello:
In this case, we have both `Please.js` and `Please.min.js`, to prevent breaking the structure, should use non-minified file here, and minify the file when auditing updates.

We can wait for https://github.com/Fooidge/PleaseJS/issues/53, if the @Fooidge likes to change the file under `dist` be `Please.min.js`, I think we could still use dist file directly.

Change the structure, breaking old-style url in new version, and let the non-minified filename with minified content, doing these things at the same time will not be a good idea.

PeterDaveHello:
Any updates?

PeterDaveHello:
Ping @Piicksarn

PR #4921, opened 2015-07-11
Add git autoupdate config in pizza

Piicksarn:


PeterDaveHello:
For `dist`, let's use `**/*` unless there already exist files we don't want to include.

Piicksarn:
Have corrected it!

PR #4945, opened 2015-07-13
add git autoupdate config in aight; update package.json

Amomo:


PeterDaveHello:
What's the reason why to fix two things into only one commit?

Amomo:
You're right, I should separate them!

PR #4950, opened 2015-07-13
add git autoupdate config in architect

Amomo:


ajax/libs/architect/package.json
```diff
@@ -4,6 +4,8 @@
   "version": "0.1.0",
   "description": "Architect is a JavaScript library built on top of Web Workers. He will manage and polyfill them workers so you don’t have to.",
   "homepage": "http://architectjs.org",
+  "authors": "Etienne Lemay",
+  "licenses": "MIT",
```
PeterDaveHello: What's the relation of these two lines and auto-update?

Amomo: Ummm...No relation to auto-update.
I should remove them.

PR #4958, opened 2015-07-13
add git autoupdate config in labjs

LeaYeh:


ajax/libs/labjs/package.json
```diff
@@ -19,5 +19,13 @@
...
       "url": "https://github.com/getify/LABjs"
     }
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/getify/LABjs.git",
+    "basePath": "",
+    "files": [
+      "LAB.min.js"
```
PeterDaveHello: I think we could use `LAB*.js` here.

PR #4970, opened 2015-07-14
Update atmosphere

Amomo:


PeterDaveHello:
Should remove 635c646, you don't need to pick the real files here.

Amomo:
done~

PeterDaveHello:
So what's the reason why to bump version here?

Amomo:
Oh, I thought I should also update that.
Okay, fixed.

PR #4979, opened 2015-07-14
[author][new] cookies-monster as minimalistic EU cookie law solution

WebReflection:
Fully customizable in both layout and optionally CSS, the bare-minimum script will take care of everything needed in order to support the European cookie law.

[Live example](http://webreflection.github.io/cookies-monster/test/test.html)

Thank you


WebReflection:
this feels like a chicken-eggs solution ... I can't explain  how simple would be to include this script in every website served in europe or for european countries untile there's a cdn for that. Should I grab my own space for it? Thanks anyway for not rejecting upfront.

PeterDaveHello:
Not really like the chicken-eggs problem, seems there is no any star/fork/watch, we'll like to serve libs, but can not audit all the codes and something else, the best way to make sure this lib been served will help people is the population, I think we don't need really high population here, just few of dozens will be fine, what do you think?

WebReflection:
fair enough, I'll try to promote it and see how the community reacts. Thanks

PeterDaveHello:
Thank you @WebReflection

WebReflection:
is 10+ stars enough? https://github.com/WebReflection/cookies-monster

PeterDaveHello:
nice

WebReflection:
:+1: thanks!

PR #4997, opened 2015-07-15
add nouislider v8.0.2 with git autoupdate

maruilian11:


PeterDaveHello:
Already have it https://cdnjs.com/libraries/noUiSlider, you should update it not re-add it duplicated, and you can also add git auto-update config for it.

PR #5006, opened 2015-07-16
Add autoupdate for html5-history-api

bdukes:


ajax/libs/html5-history-api/package.json
```diff
@@ -3,5 +3,14 @@
...
   "description": "HTML5 History API provides an emulation of HTML5 History API for older browsers",
-  "homepage": "https://github.com/devote/HTML5-History-API"
+  "homepage": "https://github.com/devote/HTML5-History-API",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/devote/HTML5-History-API.git",
+    "basePath": "",
+    "files": [
+      "history.min.js",
+      "history.ielte7.min.js"
```
PeterDaveHello: I think we should use `history*.js` here.

bdukes: Alright, I've made that update.

PR #5015, opened 2015-07-17
Added jquery-ui-i18n.min.js for internationalization

RdeWilde:
Please add


PeterDaveHello:
What's the source of this file?

RdeWilde:
Google CDN. There's another pull-request with the same file, just this one IS minified

PeterDaveHello:
Any official document or reference?

RdeWilde:
You can just Google it or compare it to the files from the Github repo;

https://github.com/jquery/jquery-ui/tree/master/ui/i18n

http://stackoverflow.com/questions/25765124/where-are-the-i18n-files-of-jquery-ui-datepicker-1-11-0

PeterDaveHello:
So this combine is from Google not from jquery-ui, am I right? I wonder if we should add it in.
cc @drewfreyling @ryankirkman @thomasdavis @IonicaBizau

ryankirkman:
@PeterDaveHello It seems like it would be useful to some people. The question is can we automate updates to this in the future, or will it have to be manual?

IonicaBizau:
@ryankirkman :+1:

PeterDaveHello:
That's the problem, seems to be manual, and if we want to do it, should set up a document or database for this special cases so that maintainers can do it correctly in future, don't know what you guys think.

PeterDaveHello:
Hello, any updates here? Thanks.

PeterDaveHello:
Any updates here? Thanks!

sufuf3:
Actually, we added this file https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/jqueryui/1.11.4/i18n/jquery-ui-i18n.min.js
@PeterDaveHello can we close this PR?
Thank you.

PeterDaveHello:
Let's close it, thanks @sufuf3

PR #5034, opened 2015-07-19
Update card configuration to auto-update from NPM

jessepollak:


jessepollak:
Thanks!

PeterDaveHello:
np

PeterDaveHello:
Oops, the config is wrong ... should not include js as basePath !

PR #5086, opened 2015-07-23
Add git autoupdate config in gumby

LeaYeh:
@maruilian11  could you plz help check this commit?

### DIFF
--- a/ajax/libs/gumby/package.json
+++ b/ajax/libs/gumby/package.json
@@ -20,5 +20,16 @@
       "type": "git",
       "url": "https://github.com/GumbyFramework/Gumby.git"
     }
-  ]
+  ],
+  "autoupdate": {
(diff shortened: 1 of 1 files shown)


maruilian11:
i think we also need img, and  it seems that we don't need to include "gumby.min.js" twice.

maruilian11:
i think it's ok. @PeterDaveHello

PeterDaveHello:
What's the reason why too grab a empty css file?
https://github.com/GumbyFramework/Gumby/blob/2.6.4/css/style.css

Do we need these all files?
https://github.com/GumbyFramework/Gumby/tree/2.6.4/js/libs

I think you should ask its author when you don't really know what files should be added.

PeterDaveHello:
Any updates?

PR #5082, opened 2015-07-23
add flux v2.0.3 with git autoupdate, close #5080

maruilian11:
@Piicksarn could u plz help me check this commit?
it's about add a new lib with git autoupdate. thx~

PR #5105, opened 2015-07-24
[new] Dispatcher for Facebook's Flux app architecture pattern

taydakov:
Flux is an app architecture designed at Facebook, its page is https://facebook.github.io/flux/, repo is https://github.com/facebook/flux
Flux works the best with ReactJS and it uses special Dispatcher that developers have to implement by themselves or use "flux" npm module.
So in order to make web explorer's life easier let's put it into the cloud ;)


PeterDaveHello:
Please squash them into only one commit, thanks.

PeterDaveHello:
Added in #5082

Issue #5137, opened 2015-07-27
[Request] Add FuckAdBlock & BlockAdBlock


sitexw:
**Library name:** FuckAdBlock & BlockAdBlock
**Git repository url:** https://github.com/sitexw/FuckAdBlock
**Git repository url:** https://github.com/sitexw/BlockAdBlock
**npm package url(optional):** https://www.npmjs.com/package/fuckadblock
**npm package url(optional):** https://www.npmjs.com/package/blockadblock
**Official homepage:** http://sitexw.fr/fuckadblock/
**Wanna say something? Leave message here:**

FuckAdBlock and BlockAdBlock are perfectly identical, except the name. For FuckAdBlock name too aggressive for certain company / website.

=====================
Notes from cdnjs maintainer:
You are welcome to add a library via sending pull request,
it'll be faster then just opening a request issue,
and please don't forget to read the guidelines for contributing, thanks!!

PR #5153, opened 2015-07-28
add blockadblock v3.1.1 with git autoupdate

maruilian11:
@Amomo could u plz help me check this pr for #5137, thx~


Amomo:
Hi @maruilian11 , the popularity of this lib seems not enough currently.
@PeterDaveHello , I am wondering that we should add it now or not.

maruilian11:
@Amomo oh i didn't notice that haha, closing is fine for me.

PeterDaveHello:
Acceptable popularity:
[sitexw/BlockAdBlock/README.md](https://github.com/sitexw/BlockAdBlock/blob/master/README.md)
> ([FuckAdBlock](https://github.com/sitexw/FuckAdBlock) same project but with a more convenient name)

But I wonder if we should host a lib to against users' rights about advertisement, I'll discuss this with other maintainers.

sitexw:
What do you think then?

PeterDaveHello:
cc @cdnjs/team-cdnjs @cdnjs/intern

Piicksarn:
@cdnjs/intern.

maruilian11:
we need to wait for the maintains' decision, don't we?

PeterDaveHello:
what do you think @cdnjs/intern ?

maruilian11:
i think user can choose if they want to use this lib, they still have the right.

PeterDaveHello:
> i think users can choose if they want to use this lib, they still have the right.
That's the reason why users use adblock/ublock, pushing a lib to block them may make them no choice.

sitexw:
> That's the reason why users use adblock/ublock

I do not agree he uses AdBlock against advertisements and not against the ad blocker detectors.

FuckAdBlock and BlockAdBlock were not designed to be against the user, but simply to allow webmasters to tell their users that they need these ads to survive.
Obviously, the other person uses to block access to content if they have an ad blocker, but it is their right, it is the internet, is open.
But do not deprive the many sites (small and large) that exist and live through internet.

PeterDaveHello:
@sitexw thanks for your comment.

PeterDaveHello:
Let's add this, please rebase this base upon the latest master branch.

PeterDaveHello:
What's going on here？

maruilian11:
@PeterDaveHello sorry, deleted the branch carelessly, new pr is here.
https://github.com/cdnjs/cdnjs/pull/7197

PR #5158, opened 2015-07-28
Add git autoupdate config in dollar.js

LeaYeh:
@Amomo  Would you mind checking this PR for me?

### DIFF
--- a/ajax/libs/dollar.js/package.json
+++ b/ajax/libs/dollar.js/package.json
@@ -23,5 +23,14 @@
     "AMD",
     "oz",
     "ozjs"
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/dexteryy/DollarJS.git",
+    "basePath": "",
+    "files": [
+      "dollar*.js",
+      "dollar/**/*"
+    ]
+  }
 }


Amomo:
Hi @LeaYeh , in latest version, the content of `dollar.js` only contains 3 lines of code.
By reading original repo's package.json, I think we should also add `dollar/` directory.

LeaYeh:
thank you pointed out my mistake which i never consider :)

Amomo:
@LeaYeh  sorry!! I am late for checking this PR.
I think we should provide every files in dollar/ to satisfy usages on different platform.

Amomo:
I think it is good now!
@PeterDaveHello could you take a look?

ajax/libs/dollar.js/package.json
```diff
@@ -23,5 +23,14 @@
...
     "ozjs"
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/dexteryy/DollarJS.git",
+    "basePath": "",
+    "files": [
+      "dollar*.js",
+      "dollar/**/*"
```
PeterDaveHello: I don't think we need it, and you should not use wildcard in non-dist folder ...

Amomo: Don't we need the files in `dollar/` ?

PeterDaveHello: Why do we need? I think this auto-update is not good at all, both line 32 and line 33.

PeterDaveHello:
You should read the content in the files you want to add.

Issue #5195, opened 2015-08-01
Point filename field to minified file


PeterDaveHello:
These is the value `filename` field without suffix `.min.` in the libs we hosted, not sure if they are already minified or not, **may even not be correct**, the filename field should point to a minified and production mainfile. for example, see this commit for js-data: 2703d3b.
- [x] browser.js
- [x] svg-morpheus.js
- [x] ace.js
- [x] com.iskitz.ajile.js
- [x] amcharts.js
- [x] ammap.js
- [x] amstock.js
- [x] angular-locale_en-us.js
- [x] audiosynth.js
- [x] aui/js/aui.js
- [x] backbone-react-component.js
- [x] backbone.modal-bundled.js
- [x] basics.css
- [x] benchmark.js
- [x] mindb
- [x] ckeditor.js
- [x] coin-slider.js
- [ ] cookiejar.js
- [x] cookies-monster.js
- [x] rollups/md5.js
- [x] load1.css
- [x] PIE_IE678.js
- [x] diff_match_patch.js
- [x] document-register-element.js
- [x] dojo.js
- [x] dom4.js
- [x] dygraph-combined.js
- [x] echarts.js
- [x] eddy.dom.js
- [x] es-class.js
- [x] ext-all.js
- [x] jquery.fancybox.pack.js
- [x] script/GeoExt.js
- [x] highcharts.js
- [x] highmaps.js
- [x] highstock.js
- [x] hprose-html5.js
- [x] ie8.js
- [x] jquery.serialize-object.compiled.js
- [ ] jquery.ui.map.js
- [x] jquery.carouFredSel.packed.js
- [x] sha.js
- [x] jsxgraphcore.js
- [x] kwargs.js
- [x] leaflet.draw.js
- [x] leaflet.markercluster.js
- [x] leaflet.js
- [ ] maps/svg/production/ic_beenhere_24px.svg
- [x] MathJax.js
- [ ] lang.js
- [x] mori.js
- [x] musicmetadata.js
- [x] OpenLayers.js
- [ ] picker.js
- [x] piwik.js
- [x] proj4.js
- [x] base.js
- [x] query-result.js
- [x] rangy-core.js
- [x] async.js
- [x] restyle.js
- [x] sbt-core-dojo-amd.js
- [x] css/scojs.css
- [x] sea.js
- [x] shaka-player.compiled.js
- [x] spf.js
- [x] system.js
- [x] tagmanager.js
- [ ] font/icomatic.css
[...]

PR #5206, opened 2015-08-01
point backbone.modal-bundled.js to backbone.modal-bundled-min.js, #5195

maruilian11:
@Amomo could u plz help me check this pr from #5195, thx.


Amomo:
@PeterDaveHello I think it is fine.

PR #5237, opened 2015-08-05
[author] jQuery-Form-Validator - Auto update should include stylsheets

victorjonsson:


PeterDaveHello:
I think we should not touch its `version` field here.

victorjonsson:
Oh... Is it okey if I do a new PR that contains both the changes to the file declaration in package.json and the missing CSS-files in the latest version (that were auto-updated)?

PeterDaveHello:
I think you can just update the filemap, and I can take the rest things.

victorjonsson:
Ok, gr8!

PR #5252, opened 2015-08-06
[author][new] Adds Freewall.js

tonytlwu:


ajax/libs/freewall/package.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "git",
+      "url": "git://github.com/kombai/freewall.git"
+    }
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/kombai/freewall.git",
+    "basePath": "",
+    "files": [
+      "freewall.js"
```
PeterDaveHello: I think you should also add a line including `"freewall.min.js"` here.

PeterDaveHello:
Would you like to add `freewall.js` in your commit? I saw you referenced it in auto-update filemap.

tonytlwu:
Actually it was meant to be freewall.min.js but the repository doesn't have a minified version.

A minified version has been added to the repository, so I'll submit this again.

PR #5255, opened 2015-08-06
[author][new] Adds Freewall.js

tonytlwu:


PeterDaveHello:
But it's 1.0.5 not 1.05 here: https://github.com/kombai/freewall/ ?

tonytlwu:
I was basing it on the file's meta data https://github.com/kombai/freewall/blob/master/freewall.js

Do you have a preference?

PeterDaveHello:
I think it should be the same as in the tag list:
https://github.com/kombai/freewall/tags

Issue #5156, opened 2015-07-28
[Request] Add flowthings


gmmeyer:
**Library name:** flowthings browser client
**Git repository url:** https://github.com/flowthings/browser-client
**npm package url(optional):** https://www.npmjs.com/package/flowthings-browser
**Official homepage:** https://flowthings.io
**Wanna say something? Leave message here:**

=====================
Notes from cdnjs maintainer:
You are welcome to add a library via sending pull request,
it'll be faster then just opening a request issue,
and please don't forget to read the guidelines for contributing, thanks!!

Piicksarn:
@gmmeyer, thanks for your suggestion！

However, README.md shows that :
```>
Libraries must have notable popularity: 100 stars or watchers on GitHub is a good example,
but as long as reasonable popularity can be demonstrated the library will be added.
```
Would you please make a request again when the library has notable popularity? :smiley:

maruilian11:
@gmmeyer thank you for your contribution, the issue will be closed in 10 days due to no response.

PR #5259, opened 2015-08-06
Adds flowthings browser and angular clients

gmmeyer:
These add both of the flowthings browser related clients (one for angular, one for plain old javascript) to the repo. I opened an issue for this,  #5156, and now am making a pr.


PeterDaveHello:
The indent of `package.json` was wrong, would you please fix it? Youcan take the already exist lib as your  example, thanks.

PR #5280, opened 2015-08-08
add min file and point audiosynth.js to audiosynth.min.js #5195

LeaYeh:


LeaYeh:
@maruilian11 Would you mind checking this PR for me?

maruilian11:
@PeterDaveHello i think it's ok.

PR #5205, opened 2015-08-01
point backbone-react-component.js to backbone-react-component-min.js

maruilian11:
@LeaYeh could u plz help me check this pr from #5195? thx.

PR #5281, opened 2015-08-08
point backbone-react-component.js to backbone-react-component-min.js #5195

LeaYeh:


PeterDaveHello:
Duplicated with #5205

Issue #5276, opened 2015-08-07
Should not point the version number to a non-stable/production version.


PeterDaveHello:
These libraries seems to be a non-stable/production version, need to check if there is updates, should not use a non-stable/production version number in package.json.
- [ ] amazeui-react version: 1.0.0-beta1 : **no stable version**
- [x] bootstrap-rtl version: 3.2.0-rc2  
- [x] css3pie version: 2.0beta1  
- [x] ember-i18n version: 3.0.0-beta.2  
- [ ] github-repo-widget version: e23d85ab8f : **no stable version**
- [x] jquery-infinitescroll version: 2.0b2.120519  
- [ ] jquery-ui-bootstrap version: 0.5pre  
- [ ] jquery-ui-map version: 3.0-rc1  
- [ ] jschannel version: 1.0.0-git-commit1-8c4f7eb : **no stable version**
- [x] json-mask version: 0.1.1-alpha  
- [x] markdown.js version: 0.6.0-beta1  
- [ ] matter-js version: 0.8.0-alpha  : **no stable version**  
- [x] onsen version: 1.2.0-beta  
- [ ] pie version: 1.0beta5  
- [ ] repo.js version: 5c0eae0f1b : **no stable version**
- [ ] skycons version: 1396634940 : **no stable version**
- [x] two.js version: 0.1.0-alpha  
- [x] unsemantic version: 0  
- [ ] web-starter-kit version: 0.2.0-beta  
- [ ] zocial version: 0

cc @Piicksarn @maruilian11 @LeaYeh @Amomo 
Please take this issue, add mention this issue ticket in each PR's comment, thanks.

LeaYeh:
okay.

Amomo:
I found that `css3pie` and `pie` are point to same repository!

PeterDaveHello:
#5287

Amomo:
Hi @PeterDaveHello ,
`jquery-infinitescroll`, `markdown.js`, `onsen`, `json-mask`, `two.js` and `unsemantic`
have pointed to a stable version and been merged so far.

And, I found some libraries haven't release a stable version yet.
- skycons
- amazeui-react
- github-repo-widget
- jschannel
- repo.js

Amomo:
There are still some libraries that I have no idea how to deal with them.
- jquery-ui-bootstrap version: 0.5pre
- ember-i18n version: 3.0.0-beta.2
- web-starter-kit version: 0.2.0-beta
[...]

PR #5296, opened 2015-08-10
add auto-update config & update to a stable version of jquery-infinitescroll

Amomo:
Hi @Piicksarn , From #5276, I would like to fix jquery-infinitescroll to point to a stable version.
repo: https://github.com/infinite-scroll/infinite-scroll
The latest stable version of jquery-infinitescroll is v2.1.0, but cdnjs doesn't have now.
So I add v2.1.0 and update the version number in package.json.
I also add git auto-update config to package.json.


Piicksarn:
i think it's fine!
@PeterDaveHello would you please take a look on the PR?
Thank you!

PR #5298, opened 2015-08-10
add auto-update config & point to a stable version of markdown.js

Amomo:
Hi @LeaYeh , From #5276, I would like to fix markdown.js to point to a stable version.
repo: https://github.com/evilstreak/markdown-js
This lib doesn't have auto-update, so I add npm auto-update config in package.json.
I think you may need to install it from npm for checking the file map.
The latest version on npm is v0.5.0,  but it doesn't provide minified version.
cdnjs has the minified file already, so I only record it in "requiredFile" field.

### DIFF
--- a/ajax/libs/markdown.js/package.json
+++ b/ajax/libs/markdown.js/package.json
@@ -20,5 +20,14 @@
       "type": "git",
       "url": "https://github.com/evilstreak/markdown-js"
     }
+  ],
+  "npmName": "markdown",
+  "npmFileMap": [
+    {
+      "basePath": "lib/",
+      "files": [
+        "markdown*.js"
+      ]
+    }
   ]
 }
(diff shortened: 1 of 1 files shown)


LeaYeh:
i think it is fine @PeterDaveHello

PeterDaveHello:
No need to use `requiredFile` here.

Amomo:
Oh! You are right! The min file has been noted in filename field.
@LeaYeh I have corrected it.
Could you come to check it again? Thank you~

Amomo:
Oh! You are right! The min file has been noted in filename field.
@LeaYeh I have corrected it.
Could you come to check it again? Thank you~

LeaYeh:
fine :) @PeterDaveHello

PR #5441, opened 2015-08-19
[author] Use autoupdate for oCanvas

koggdal:
Built files are now in the git repo and it would be great to have it auto update from there. Thanks!


PeterDaveHello:
Thanks @koggdal.

PR #5456, opened 2015-08-20
[new] flot.tooltip 0.8.5

dottedmag:
[x] new library
[x] new issue/PR
[x] popularity
[x] github
[x] tags
[x] autoupdate
[x] autoupdate target/source valid
[x] filemap is correct

Files are taken from https://github.com/krzysu/flot.tooltip/tree/0.8.5/js


ajax/libs/flot.tooltip/package.json
```diff
@@ -0,0 +1,28 @@
...
+      "url": "https://github.com/krzysu/flot.tooltip"
+    }
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/krzysu/flot.tooltip.git",
+    "basePath": "js",
+    "files": [
+      "jquery.flot.tooltip.js",
+      "jquery.flot.tooltip.min.js"
```
PeterDaveHello: I think we could use `jquery.flot*.js`?

dottedmag: In the same directory there is

* a bunch of unneeded files (at least for 0.8.5, it was cleaned later), which are unversioned copies of Flot library at some point of time
* `jquery.flot.tooltip.source.js`, that is a `jquery.flot.tooltip.js` without attribution (which is added by some build script).

PeterDaveHello:
@cdnjs/intern please review this PR.

Amomo:
@PeterDaveHello As dottedmag said, `jquery.flot.tooltip.js` only has more lines in comment field than `jquery.flot.tooltip.source.js`, they have same content.

I think this PR is ok, could you check it again? Thank you :-D

PR #5537, opened 2015-08-28
[author] Add KaTeX Autoupdate

xymostech:
This adds an autoupdate section to the KaTeX package.json.

Some questions:
 - I didn't actually add the current version of KaTeX nor the intermediate versions which we have missed updating; do you want me to add the current one here?
 - Does the autoupdating look at the git tags from a repository? We have recently started actually committing built files in separate off-master commits that are tagged (which is why we didn't add auto-updating before!). Will that work? This also means that some older versions won't be able to be auto-updated, since the built files don't exist in the repository at those tags. Do you want me to add those manually?

Hopefully this works. Thanks!

### DIFF
--- a/ajax/libs/KaTeX/package.json
+++ b/ajax/libs/KaTeX/package.json
@@ -14,5 +14,16 @@
   "repository": {
     "type": "git",
     "url": "git://github.com/Khan/KaTeX.git"
+  },
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/Khan/KaTeX.git",
+    "basePath": "dist",
+    "files": [
+      "katex.min.js",
+      "katex.min.css",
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
Would you please squash them into only one commit as our document said? Thanks.

xymostech:
I squashed the commits.

xymostech:
(any update on this?)

PeterDaveHello:
Oops, sorry, I missed this PR, would you mind to rebase with our latest master branch? Or I can manually pick this commit, thanks!

xymostech:
Okay, I rebased it again. :) You move fast!

PeterDaveHello:
Thank you @xymostech

Issue #5500, opened 2015-08-25
Add missing license field in libs' package.json 


PeterDaveHello:
cc #5194, https://github.com/cdnjs/new-website/issues/66

Should be care of the libs with multiple licenses, especially the libs have commercial license. 
 - [ ] 1140
 - [x] across-tabs
 - [ ] animations
 - [ ] boexfi
 - [x] browser-logos
 - [ ] chrome-frame
 - [x] cryptico
 - [x] davis.js
 - [x] DinaKit
 - [x] documentup
 - [x] entypo
 - [x] expect.js
 - [x] flipCounter
 - [x] foundicons
 - [x] galleriffic
 - [x] gsap
 - [ ] jcalculator
 - [x] jquery-hashchange
 - [x] jquery-outside-events
 - [x] jquery-parallax
 - [x] jquery-sparklines
 - [x] jquery-tools
 - [x] jquery.activity-indicator
 - [x] jquery.formalize
 - [x] jquery.lifestream
 - [x] jquery.percentageloader
 - [x] jquery.postcodify
 - [x] jquery.selectboxit
 - [x] jquery.socialshareprivacy
 - [x] jquery.tiptip
 - [x] json2
 - [x] legojs
 - [x] linq.js
 - [x] log4javascript
 - [ ] luminateExtend
 - [x] meyer-reset
 - [ ] mojio-js
 - [x] openajax-hub
 - [x] PKI.js
 - [x] polyglot
 - [x] prettydiff
 - [ ] probtn
 - [x] prototype
 - [x] protovis
 - [x] react-chartjs-2
 - [x] repo.js
 - [x] require-cs
 - [x] require-jquery
 - [x] roundabout
 - [x] salesforce-canvas
 - [x] script.js
 - [x] simplecartjs
 - [x] skel-layers
 - [x] skycons
 - [x] slabText
 - [x] thorax
 - [x] Vague.js
 - [ ] webicons

Amomo:
update the list according #5509 #5484

PeterDaveHello:
:+1:

LeaYeh:
update the list according #5561 #5560 #5559 #5558 #5582 #5583

Amomo:
@LeaYeh you couldn't mark the libraries if its modified PR hasn't been merged! 😫

PeterDaveHello:
@Amomo is correct ...

LeaYeh:
RRRRRRRR
sorry for that.

[...]

PR #5575, opened 2015-08-31
Enrich backbone-localstorage.js's package.json(field: license)

bmiwdmb:
#5500


PeterDaveHello:
Great, it'll be even better if you can provide link like this: https://github.com/jeromegn/Backbone.localStorage#license to prove your change, so that we can audit PR faster.

Issue #5527, opened 2015-08-27
git auto update for marked


smarts:
I wan to open a pull request to update [marked](//github.com/chjj/marked), but right now cdnjs is several versions behind the current version. For example, let's say the version in cdnjs is 5, and the latest version is 9. If I add autoupdate *and* change the package version to 6, will auto-update pull in version 6 and then continue on to pull 7, 8, and 9 into cdnjs? That's what I would like to happen. Or would I need to manually pull in all the versions that are not in cdnjs and then configure auto update?

maruilian11:
@smarts yes, it will autoupdate to 9 if they all be tagged. the package.json will update to the latest one, so welcome to send the pull request.

smarts:
@PeterDaveHello volunteered to do this.

PR #5578, opened 2015-08-31
Add autoupdate section for marked

smarts:
Resolves cdnjs/cdnjs#5527

### DIFF
--- a/ajax/libs/marked/package.json
+++ b/ajax/libs/marked/package.json
@@ -18,5 +18,14 @@
     "markdown",
     "markup",
     "html"
-  ]
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/chjj/marked.git",
+    "basePath": "",
+    "files": [
+      "marked.min.js",
+      "lib/marked.js"
+    ]
+  }
 }


PeterDaveHello:
To keep the origin structure on cdnjs, I think you should set `basePath` to `lib` and that the auto-updater grab un-minified source only, the minified file will be added when I commit new updates. Thanks.

smarts:
Can you specify exactly what you want? I'm new to this and I'm not quite sure what you mean. The documentation is much clearer for npm autoupdate sources than it is for a git source…

PeterDaveHello:
As I said:
 * set `basePath` to `lib`
 * only put `marked.js` in the `files` array.
Thanks.

smarts:
Then how will `marked.min.js` make it to the CDN?

PeterDaveHello:
I'll do it.
https://github.com/cdnjs/cdnjs#enabling-gitrecommended-or-npm-auto-update

smarts:
Yeah i already read that. And I maintain that the npm case is better documented. It tells about how it looks in the tarball and what is produced. That info is not detailed for the git case. And you didn't answer the question about how `marked.min.js` would end up as part of what gets CDN-ized.

PeterDaveHello:
I think it said we'll handle the minify works :)
> ... and we may need to minify the lib without pre-minified dist files ...

smarts:
Yikes. OK, that's good to know. That actually makes me want to avoid using cdnjs since it means I would potentially get something different than what the lib author has released on GitHub. Anyway, thanks for volunteering to do this. I'll go ahead and close this pull request.

PeterDaveHello:
We'll support multi-filemap soon, that means the files don't need to be the same basePath, which can include the minified files and src files at the same time from source.

PR #5616, opened 2015-09-03
[new] Add object-fit polyfill

peteygao:
The original source is from:
https://github.com/anselmh/object-fit


PeterDaveHello:
Nice! You can try to use `**/*` in the filemap next time 'cause the basePath is a dist folder!

PR #5657, opened 2015-09-07
[author] Updating GSAP to 1.18.0

jackdoyle:
We always seem to have trouble updating the /latest/ directory. Would you mind helping to make sure that happens? We already published /1.18.0/ to NPMJS so hopefully that'll automatically get picked up, but I don't think the /latest/ directory does.

Thanks!


ajax/libs/gsap/1.18.0/easing/EasePack.min.js
```diff
@@ -0,0 +1,12 @@
+/*!
+ * VERSION: beta 1.15.2
```
ryankirkman: @jackdoyle Did you upload a testing version here? The cdnjs tag says `1.18.0` and this says `1.15.2` :)

jackdoyle:
Yep, that's totally fine. Not every file gets updated with every release - the EasePack hasn't needed changes in a while, that's all. Thanks for even noticing and asking about this. And thanks for the quick merge! I still see that /latest/ isn't updated yet and maybe there needs to be some flushing of a cache somewhere.

ryankirkman:
Yep, emailed the Cloudflare lads about a cache purge on `gsap/latest` :)

/cc @terinjokes

PR #5681, opened 2015-09-11
added jquery-lang

profenter:
added https://github.com/Irrelon/jquery-lang-js


PeterDaveHello:
@profenter sorry for missing this PR and the late reply, we've added this lib in #6696 foe a while, you can access it via https://cdnjs.com/libraries/jquery-lang-js, thanks for your contribution!

PR #5721, opened 2015-09-17
[author] [new] ProStyle animation library

GaryChamberlain:
Hopefully I got this right now.  @PeterDaveHello thank-you for your help and your generous offer to cherry pick.  Yes, please cherry pick if needed.


PeterDaveHello:
A small problem here, you use `**/*.js` in auto-update config, but only add `min.js` files in the commit, I think you should make them the same, thanks.

GaryChamberlain:
I'll fix and squash.

PeterDaveHello:
hmmmm and the build failed ...

PeterDaveHello:
by the way, you can use `--amend` with commit so that you don't need to squash!

GaryChamberlain:
@PeterDaveHello Is this good?

PeterDaveHello:
I think you should also clean up the commit message, there are some unrelated strings.

GaryChamberlain:
@PeterDaveHello good idea. I just amended it. Thanks for all the help!

Issue #5730, opened 2015-09-19
gmaps.js is serving an empty link


dinolatoga:
https://cdnjs.cloudflare.com/ajax/libs/gmaps.js/0.4.19/gmaps.min.js does not seem to return anything.

PeterDaveHello:
hmmmm ... my bad ... I'm the murderer lol ...
dd35f87b191bb9796f50a00ec612cb0ea4a5b75c

PeterDaveHello:
fixed in #5736

PeterDaveHello:
@dinolatoga please confirm, and thanks for your report.

@Amomo don't forget to add `fix #5730` next time.

Amomo:
ok, I will remember that.

dinolatoga:
@PeterDaveHello the link still returns an empty file

PeterDaveHello:
@terinjokes would you please help us flush the cache of https://cdnjs.cloudflare.com/ajax/libs/gmaps.js/0.4.19/gmaps.min.js? Thanks!

dinolatoga:
The minified link is empty but the unminified file is loading properly. I'm currently using the unminified version but it would be good to have the minified version working.

PeterDaveHello:
@dinolatoga sorry for the late reply, do you still have this problem? Looks good for me now.

PeterDaveHello:
Close due to no response and the problem should be fixed.

PR #5736, opened 2015-09-19
add missing files & basePath field in gmaps.js

Amomo:
Hi @Piicksarn 

#5730 reported that `gmaps.js/0.4.19/gmaps.min.js ` is empty.
So, I add the real "gmaps.min.js" and "gmaps.min.js.map" in v0.4.19.

And, I found that in c2169d5, the author lost the `"basePath"` field.
So I add on the missing field.


PeterDaveHello:
Nice catch!

Piicksarn:
:+1:

PR #5786, opened 2015-09-28
Change React auto-updating to use git

zpao:
In facebook/react#4901 we're going to be moving `react-dom.js` outside of the `react` npm package. For now we'd like to keep that file still in the same project here on the CDN (and we're still distributing it with the `react` bower package). In order to do that we need to point the autoupdater at the repo we use for bower instead of npm.

I think the only difference here is that we store RCs here. We always published those to npm as well but because we never used the `latest` tag, they never ended up here.


PeterDaveHello:
Thanks @zpao

PR #5828, opened 2015-10-04
Enrich backbone-forms's package.json(field: license)

LeaYeh:
Hi @Amomo  , 
For #5500 , I update the license.

backbone-forms use `"MIT License"`,
You can find this info from https://github.com/powmedia/backbone-forms/blob/master/LICENSE
repo : https://github.com/powmedia/backbone-forms

- [ ] Not found on cdnjs repo
- [x] No already exist issue and PR
- [x] Notable popularity : **Watch `104`, Star `2136`, Fork `423`**
- [x] Project has public repository on famous online hosting platform (or been hosted on npm) 
- [x] Has valid tags for each versions (for git auto-update)
- [x] Auto-update setup
- [x] Auto-update target/source is valid.
- [x] Auto-update filemap is correct.

Would you mind checking this PR for me? Thank you~ :-)

PR #5919, opened 2015-10-14
Enrich 6to5's package.json(field: license)

LeaYeh:
Hi @Amomo  , 
From #5500, I have added license info. https://github.com/babel/babel/blob/master/LICENSE#L3
repo : https://github.com/babel/babel
Would you mind checking this PR for me? Thank you.

### DIFF
--- a/ajax/libs/6to5/package.json
+++ b/ajax/libs/6to5/package.json
@@ -33,5 +33,6 @@
         "browser.js"
       ]
     }
-  ]
+  ],
+  "license": "MIT"
 }


Amomo:
Hi @LeaYeh as Peter said in #5828

You should make your commit message easy to recognize `which license you add`.
Don't use only `field: license`.

And, your PR comment shouldn't be written in the commit message.

LeaYeh:
i think @PeterDaveHello means we should point out which line in file we recognized that is "MIT" not about commit.
https://github.com/babel/babel/blob/master/LICENSE#L3

maruilian11:
i thought he said `LICENSE` **file** is not so reliable.

LeaYeh:
@maruilian11  but it is clearly point out that is "MIT", or what do you mean **reliable** ?

PeterDaveHello:
If you only point to a file, the file may not be so reliable, unless you can tell me every time you did that you really read all the sentences and words in it, if not, I think there should be a word "MIT" or something to clarify its license, that's all, in this case, line 3 already said its MIT, it's okay, isn't it?

maruilian11:
oh got it

Amomo:
Oh, I see the difference! Thank you everyone~

I am sorry @LeaYeh,
This is a good PR, could you rebase it? Thank you! :-D

LeaYeh:
@Amomo ok, i already rebased it. pls check it again

Amomo:
@PeterDaveHello I think this PR is fine, could you take a look?
Thank you! :-)

PR #5954, opened 2015-10-17
[Author] Updating Fluidbox package.json

terrymun:
Thanks for the hard work, guys! +1 

I have recently updated Fluidbox to v2, which comes with changes to the directory structure as the files are now moved to the dist/ directory. This is reflected in the basePath property.

Please let me know if anything else needs to be looked into.

PR #5965, opened 2015-10-19
Updated Fluidbox

terrymun:
Removed the merge commit from the master branch (should've used rebase, but totally forgot). So now there is only one commit and should be a clean PR.

Thanks guys! :+1:


PeterDaveHello:
cc #5954

terrymun:
@PeterDaveHello Thanks Peter :)

PeterDaveHello:
@terrymun Thank you too.

PeterDaveHello:
@terrymun I just found that the structure was wrong. I think you should put the css and its map file under a css folder and so does js file, just like the structure in your dist folder.

terrymun:
Alright. Should I submit another PR then? Just to modify package.json and add the map files to the respective folders.

PeterDaveHello:
I just add the correct structure in, and leave the wrong one alone to prevent breaking current usage. There will be only correct one in next release.

PeterDaveHello:
Also updated the filemap for this lib:
https://github.com/cdnjs/cdnjs/commit/566381137c55c22b2ef7bd7d4c8f3283838d0120
```diff
@@ -34,8 +34,7 @@
     "target": "git://github.com/terrymun/Fluidbox.git",
     "basePath": "dist",
     "files": [
-      "**/*.min.js",
-      "**/*.min.css"
+      "**/*"
     ]
   },
   "repository": {
```

terrymun:
Thanks Peter! But does that mean I can't make a PR to pull the map files in?

PeterDaveHello:
I think I added, so why do you need to add it?

terrymun:
But the map files are not in the PR I made... or would that be updated automatically?

PeterDaveHello:
I added in the right place:　https://github.com/cdnjs/cdnjs/commit/65645474156e1bec59a43c3368ba522fb47fa951
For the wrong place, I think we can just leave it alone.

PR #5968, opened 2015-10-19
[author] Updated package.json for remoteStorage

gunar:
Added auto-update tag and made `package.json` compliant to `test/schemata/npm-package.json`.

FYI, previously [remotestorage/remotestorage.js](https://github.com/remotestorage/remotestorage.js) directory structure was

    release/v0.8.0
    release/v0.9.0
    ...

now it's just

    release/stable

.

Thanks a bunch.

### DIFF
--- a/ajax/libs/remoteStorage/package.json
+++ b/ajax/libs/remoteStorage/package.json
@@ -4,12 +4,6 @@
   "filename": "remotestorage.min.js",
   "description": "Client-side Javascript library to make apps remoteStorage-compatible.",
   "homepage": "http://remotestorage.io/",
-  "maintainers": [
-    {
-      "name": "Niklas Cathor",
-      "url": "nil.niklas@googlemail.com"
-    }
-  ],
   "licenses": [
     {
       "type": "AGPL",
@@ -17,6 +11,7 @@
     },
     "MIT"
   ],
+  "license": "MIT",
   "keywords": [
     "remoteStorage",
     "unhosted",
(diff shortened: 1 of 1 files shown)


gunar:
`test/schemata/npm-package.json` doesn't describe properties "maintainers", "license" or "repository" of `package.json`, so I've removed them. 

If that's not correct please let me know.

PeterDaveHello:
@gunar sorry but please add license info back, the reason why it's not in the schema is that we are still confirming all the libs' license. #5500

gunar:
@PeterDaveHello remoteStorage can be used under the AGPL license as well. Can I add the "license**s**" field back too?

PeterDaveHello:
I think we can use only `licenses` as array to store them.

gunar:
@PeterDaveHello Ok so if this is compliant, then I'm happy with it :) Thanks!

PeterDaveHello:
One last thing, would you please squash these commits as our document says into only one commit? Thanks.

gunar:
@PeterDaveHello I think this is it. Thank you!

PeterDaveHello:
Thanks, I'll clean up the license part later, merge now!

PeterDaveHello:
I just clean it up in d17d756 @gunar, but I only saw "MIT" in its package.json here -> https://github.com/remotestorage/remotestorage.js/blob/master/package.json

So maybe you should also update it? cc @remotestorage

gunar:
@PeterDaveHello Surely. Thank you very much! :grin:

PR #6042, opened 2015-10-26
Add FieldKit to CDN JS

iamJoeTaylor:
Please let me know if I'm doing this incorrectly.

Thanks,
-Joe


PeterDaveHello:
Hi @iamJoeTaylor, would you please squash them into only one commit and use much meaningful commit message, thanks.

cc @cdnjs/intern please also help review this PR.

ajax/libs/field-kit/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "npmName": "field-kit",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "*.js"
+      ]
+    }
+  ],
+  "repository": "square/field-kit"
```
PeterDaveHello: This is not valid for us, should use a object to say its type and url.

ajax/libs/field-kit/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "field-kit",
+  "filename": "field-kit.min.js",
+  "version": "2.0.4",
+  "description": "Utilities to effectively manage text entry on the web.",
+  "homepage": "http://square.github.io/field-kit/",
```
PeterDaveHello: I'll respectfully suggest you to use https here.

ajax/libs/field-kit/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "format"
+  ],
+  "author": "Square, Inc.",
+  "license": "Apache-2.0",
+  "npmName": "field-kit",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "*.js"
```
LeaYeh: "*.+(js|map)"
@iamJoeTaylor i think we also need `field-kit.min.js.map` and `field-kit.js.map`

iamJoeTaylor:
Thanks for all the feedback. I think I've addressed all the issues.

ajax/libs/field-kit/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "format"
+  ],
+  "author": "Square, Inc.",
+  "license": "Apache-2.0",
+  "npmName": "field-kit",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "*.+(js|map)"
```
PeterDaveHello: I think use "**/*" will be even better, what do you think?

PeterDaveHello:
And it'll be even better if you can change the commit message from "add field-kit lib" to "add field-kit lib v2.0.4", thanks!

iamJoeTaylor:
@PeterDaveHello done

PeterDaveHello:
Thanks @iamJoeTaylor

PR #6058, opened 2015-10-28
add history 1.12.6

ashwinr:


ajax/libs/history/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "location"
+  ],
+  "filename": "History.min.js",
+  "npmName": "history",
+  "npmFileMap": {
+    "basePath": "umd",
+    "files": [
+      "*.js"
+    ]
+  }
```
PeterDaveHello: Should be an array here, fixed in 08060042099c3b4aa6ccee04fea3b9c307333378

PR #6081, opened 2015-10-31
Add typeahead-addresspicker@0.1.4 with git autoupdate

steve0hh:
**Library name: bootstrap-progressbar**
**Git repository url: https://github.com/sgruhier/typeahead-addresspicker**
**License(s): The MIT License (MIT) [found here](https://github.com/sgruhier/typeahead-addresspicker/blob/master/LICENSE)**
**Official homepage: http://sgruhier.github.io/typeahead-addresspicker/**

cc @sgruhier


sgruhier:
What is git autoupdate ?

steve0hh:
@sgruhier 
> cdnjs automatically updates libraries that are known to be hosted on npm or git repo, e.g., Lodash. npm auto-update relies on each release and git auto-update relies on the tags in git repo. This auto-update script runs every hour, but the update result won't be committed until one of our maintainers audit the updates, because many libs will change the naming or directory structure during different versions, and we may need to minify the lib without pre-minified dist files, so it'll be reasonable to delay for at most 30 hours, if you think there is a lib didn't been updated, please wait at least 30 hours for the process to audit it.

PeterDaveHello:
Please also add license field in package.json, thanks.

ajax/libs/typeahead-addresspicker/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "google",
+    "map"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/sgruhier/typeahead-addresspicker.git",
+    "basePath": "dist",
+    "files": [
+      "typeahead-addresspicker.js",
+      "typeahead-addresspicker-min.js"
```
PeterDaveHello: Should use `"**/*"` instead of the two lines above.

steve0hh:
add license field and change files to `"**/*"`. :smile:

PeterDaveHello:
Thanks!

PR #6082, opened 2015-10-31
update lazy to 3.0.1101beta

keicheng:


keicheng:
sorry , i redo, this fork maybe too old.

PR #6119, opened 2015-11-03
Update shaka-player package.json to include new build variants

joeyparrish:
Shaka Player v1.6.0 includes new build variants that should be distributed on cdnjs.

See also google/shaka-player#116


PeterDaveHello:
Thanks @joeyparrish

PR #6140, opened 2015-11-06
[new] browser-i18n

taufique71:
Added new library named browser-i18n which I couldn't find in https://cdnjs.com. I am not the author of the library but the library came handy for me so I thought it should be on cdnjs.

Author: Renato Gamma <renatoargh@gmail.com>
Source: https://github.com/gammasoft/browser-i18n


taufique71:
@LeaYeh Hi. Could you manage time to review the pull request. Is there any problem I need to solve?

ajax/libs/browser-i18n/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "type": "git",
+    "url": "https://github.com/gammasoft/browser-i18n.git"
+  },
+  "keywords": [
+    "i18n",
+    "browser"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/gammasoft/browser-i18n.git",
```
LeaYeh: here you need to add `basepath` info,
in this case, like this `"basePath": "", `

ajax/libs/browser-i18n/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "files": [
+      "**/*"
```
LeaYeh: and here we dont need `**/*` all files to auto update,
you can use this `i18n*.js`

LeaYeh:
hi @taufique71 besides previous problems here is main problem that  this lib doesn't have enough popularity(100 watch or star is a good example). Please take a look at our README.md for detail.
we suggest you open a issue again when the repo has notable popularity. 
thank you for your contribution.

taufique71:
Thank you for your feedback @LeaYeh :)

taufique71:
Hi @LeaYeh can you clarify a bit more? When the library has notable popularity whould I have to create a new issue or new pull request? Or this pull request will be open and I would have to give a knock? Which one will be the most appropriate procedure?

LeaYeh:
you should close it first and when the popularity is enough then send a new pr

PR #6189, opened 2015-11-12
Update package.json to include *.map files

OlegKi:
The next version of free jqGrid will provide *.map to all *.js and *.css files.

### DIFF
--- a/ajax/libs/free-jqgrid/package.json
+++ b/ajax/libs/free-jqgrid/package.json
@@ -5,12 +5,13 @@
     {
       "basePath": "/",
       "files": [
-        "css/ui.jqgrid.css",
+        "css/*",
         "js/jquery.jqgrid.min.js",
         "js/jquery.jqgrid.min.map",
         "js/jquery.jqgrid.src.js",
         "js/i18n/*.js",
-        "plugins/*.*"
+        "js/i18n/*.map",
+        "plugins/*"
       ]
     }
   ],


PeterDaveHello:
@OlegKi is it possible to setup a `dist` folder for each releases?

OlegKi:
@PeterDaveHello: Yes, I create always `dist` folder for each releases. The problem is only that the folder is inside of `.gitignore` and so the folder will be not copied to GitHub. Do you have some suggestion to change file structure of free jqGrid, which could be helpful for `cdnjs`?

PeterDaveHello:
not sure if you would like to publish npm package or push dist folder for the releases, or if you wanna just use current structure, it's also fine, but just need to update the config every time you want to change the structure from your side.

OlegKi:
Sorry. I will of cause publish new version of the package to npm, bower and other repositories. The reason of my pull request was the following: I was not sure **when** cdnjs use `package.json` - before or after collecting the data to publishing on cdnjs. I though that cdnjs could uses *the current `package.json`* from `ajax/libs/free-jqgrid/package.json` to get the files, which need be placed to CDN. If cdnjs would *first* refresh `package.json` and *then* get files for CDN, then my pull request would be not required.

In any way, *thanks you very much* for merging the pull request and for working for cdnjs in general. cdnjs is very helpful.

PeterDaveHello:
@OlegKi I am afraid that we can not refresh package.json automatically or periodically, it may cause problems and break the structure we need, so webuse the current config under our project. Cdnjs uses package.json at the time we add a lib or update a lib only, including meta data like keywords and license info.

Issue #5491, opened 2015-08-24
[Request] Auto-update config for datatables-colvis


PeterDaveHello:
Repo url: https://github.com/DataTables/ColVis

PR #6275, opened 2015-11-23
add auto-update config in datatables-colvis

Amomo:
Hi @maruilian11 

For #5491, I add datatables-colvis at v1.1.2 and add git auto-update config in its package.json.
repo: https://github.com/DataTables/ColVis

The old filename field in package.json was point to `datatables.colvis.min.js`, but datatables-colvis's repo use `dataTables.colVis.js` for the real file's name.
So I add the latest stable "version 1.1.2" to update the "filename field".

Would you mind checking this PR for me? Thank you~ :-D

- [ ] Not found on cdnjs repo
- [x] No already exist issue and PR
- [ ] Notable popularity : **Watch `10`, Star `34`, Fork `45`**
- [x] Project has public repository on famous online hosting platform (or been hosted on npm) 
- [x] Has valid tags for each versions (for git auto-update)
- [x] Auto-update setup
- [x] Auto-update target/source is valid.
- [x] Auto-update filemap is correct.


maruilian11:
@Amomo we won't add a version when we already have a lib on cdnjs, so i think you can just sent a commit to modify it's filename and its folder name.

PeterDaveHello:
hm, special case for filename changing, because there is a requiredFiles array, I would like to merge it for good reason, I think you should add the config first and then modify the filename, and please also add v1.1.1 for this case, thanks.

Amomo:
Thanks for your comment! :-)

@maruilian11 would you mind to check this PR again? thank you!

Issue #6236, opened 2015-11-18
Merge CONTRIBUTING-WIP.md into CONTRIBUTING.md


PeterDaveHello:
[CONTRIBUTING-WIP.md](https://github.com/cdnjs/cdnjs/blob/master/CONTRIBUTING-WIP.md)
[CONTRIBUTING.md](https://github.com/cdnjs/cdnjs/blob/master/CONTRIBUTING.md)

it's time to merge(maybe also need to update) these two files :+1: 

Ping me if there is any ambiguous policies or anything.

MattIPv4:
cc @AmNotADev

MattIPv4:
I did not GitHub...

MattIPv4:
I think this one probably can actually be closed, the docs in this repo will be replaced by new ones in the packages-migration repo.

xtuc:
I was concerned about the time the migration could take, maybe we want to fix this before migrating the docs over?

MattIPv4:
It seems like this would just create more work as the docs would need adjusting to be used in the current setup, and then further changes need to be made to them to reflect the new post-migration setup.

IMO, the docs in this repo have been as they are for a long time, a few more weeks/months like this before we do the migration would be fine?

PeterDaveHello:
It's no need to migrate and rewrite the doc before the new model of how the library update would work, the effort won't worth it. This one should be closed.

xtuc:
The new model of libray update will be similar enough I suppose, but I agree with @MattIPv4 that a couple weeks more  as is would be fine

PR #6304, opened 2015-11-25
modify npm auto-update example

maruilian11:
i think `dist` and `**/*` is better.

### DIFF
--- a/CONTRIBUTING-WIP.md
+++ b/CONTRIBUTING-WIP.md
@@ -36,46 +36,43 @@ Each cdnjs library has a `package.json` file. This file contains required and so
 ### 2.3 Auto-update example
 
 ```js
-  "npmName": "lodash",
+  "npmName": "function-plot",
   "npmFileMap": [
     {
-      "basePath": "/dist/",
+      "basePath": "dist",
       "files": [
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
There is no `dist` now ...

```
array/    chain/    collection/    date/    function/    index.js   lang/    LICENSE  math.js  number.js  object.js     README.md  string.js   utility/
array.js  chain.js  collection.js  date.js  function.js  internal/  lang.js  math/    number/  object/    package.json  string/    support.js  utility.js
```

maruilian11:
but according to the structure example under npm auto-update example, i think we still can use `dist` and `**/*` to let people understand what we mean. or we can change a lib name?

PeterDaveHello:
there is no dist now ... don't do a work not help

PeterDaveHello:
I think you can choose another lib as example

maruilian11:
i have modified lib to `function-plot`.

PeterDaveHello:
cc #6236

PeterDaveHello:
ping @maruilian11 to open a new issue for readme.md

maruilian11:
ok.

PR #6307, opened 2015-11-25
Add autoupdate config in froala-editor

stefanneculai:
Hello,

I have added autoupdate in package.json and fixed the Froala Editor links and licenses.

PR #6313, opened 2015-11-26
Add autoupdate config in froala-editor 

stefanneculai:
For #6307.

### DIFF
--- a/ajax/libs/froala-editor/package.json
+++ b/ajax/libs/froala-editor/package.json
@@ -1,55 +1,38 @@
 {
   "name": "froala-editor",
   "filename": "js/froala_editor.min.js",
-  "npmName": "froala-editor",
-  "npmFileMap": [
-    {
-      "basePath": "/",
-      "files": [
-        "js/plugins/*.min.js",
-        "js/froala_editor.min.js",
-        "js/froala_editor_ie8.min.js",
-        "css/froala_editor.min.css",
-        "css/froala_page.min.css",
-        "css/themes/*.min.css"
-      ]
-    }
-  ],
   "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
Sorry @stefanneculai , would you please mention the lib name in the first line of your commit message for that we can easily trace the changes from commit log? Thanks.

stefanneculai:
@PeterDaveHello Done.

PeterDaveHello:
@stefanneculai do you need files under `plugins`, `themes` in the css folder?
If so, we need to update the auto-update config.

stefanneculai:
Yes. Should `js/plugins/*` and so on be added?

PeterDaveHello:
if so, you should use `css/**/*` instead of `css/*` to grab files recursively, but in fact we usually don't use this kind of wildcard under a non-dist folder or repository, do you have any suggestions? but if you still wanna use that, I'll accept your PR as you are the author of the lib.

stefanneculai:
I have made the update. Actually we use the whole repository as a dist folder. We are doing the development of the library only internal and on Github we make public only the dist of the package.

ajax/libs/froala-editor/package.json
```diff
@@ -1,59 +1,38 @@
...
+      "type": "Advanced",
+      "url": "https://www.froala.com/wysiwyg-editor/pricing"
     },
     {
       "type": "OEM",
-      "url": "http://editor.froala.com/download"
+      "url": "https://www.froala.com/wysiwyg-editor/pricing"
     }
-  ],
-  "repository": {
```
PeterDaveHello: Can we reserve repository info?

stefanneculai: Added it back.

PeterDaveHello:
Thanks!

Issue #6324, opened 2015-11-29
Add missing license field in libs' package.json (11/29 workshop@NCTU)


Amomo:
cc #5500  

**Please ref this issue in your PR comment !**
Attention to  your `branch name` and `license reference`.

This issue is for workshop in NCTU (11/29).

These libs don't have `license` field, should add missing info in there package.json to help us maintain these libs.

- [x] ResponsiveSlides.js
- [x] Uniform.js
- [x] angular-route-segment
- [x] angulartics
- [ ] arbor
- [x] architect
- [x] atmosphere
- [x] basket.js
- [x] batman.js
- [x] bean
- [x] bitcoinjs-lib
- [x] bonsai
- [x] bootstrap-social
- [x] brain
- [x] chaplin
- [ ] commandz
- [x] cookiejar
- [x] crossfilter
- [ ] cubism
- [ ] davis.js
- [ ] depot
- [ ] dinqyjs
- [ ] dragdealer
- [ ] dragonbones
- [ ] easy-pie-chart
- [ ] easyXDM
- [x] eve.js
- [x]  fiber
- [x] fingerprintjs
- [x] gumby

PeterDaveHello:
We merged PRs from @coldmanck @adwin5 @pvnr0082t @JalexChang @fishballLin @a1090200 @Kamigami55, nice :facepunch: !

Amomo:
Hi @coldmanck @adwin5 @hsiaohsuan1l1l @pvnr0082t @JalexChang @fishballLin @a1090200 @Kamigami55 @Asu4ni @limitimil @asdf168ghjkl 
Thank you for your PR helping us maintain these libs. :smile: 

However, some libraries in this issue still don't have license info.
Would you mind helping us again? :blush:

A simple PR is welcome. 
But, it's a good chance to try using git commands to do this work! :muscle:
Hope you would give it a try. Thank you very much!

PeterDaveHello:
Since you guys forked cdnjs repository about 2 weeks ago, without delete and re-fork this repository, you will learn the concept of "remote" in git, and then sync a repository between 2 different remotes.

PeterDaveHello:
@Amomo hey let's close this issue ASAP, there are some sub issues still opened.

PR #6331, opened 2015-11-29
add license info of brain #6324

Kamigami55:
license ref: https://github.com/harthur/brain/blob/master/LICENSE


PeterDaveHello:
Will be better to separate the reference issue, like:
 - add license info of brain**,** #6324

PR #6337, opened 2015-11-29
add missing license of bean

a1090200:
#6324
add missing license of bean
[License](https://github.com/fat/bean/blob/master/LICENSE)


PeterDaveHello:
:+1:

Issue #6317, opened 2015-11-27
[Request] Add react-highcharts


kirjs:
**Library name:** react-highcharts
**Git repository url:** https://github.com/kirjs/react-highcharts
**npm package url(optional):** https://www.npmjs.com/package/react-highcharts
**License(s):** MIT
**Official homepage:** https://github.com/kirjs/react-highcharts
**Wanna say something? Leave message here:**
cdnjs is awesome, thanks a lot!

PeterDaveHello:
What are the files should be hosted on the cdn?

kirjs:
Hi peter, please see the paths below:
```bash
dist/highcharts.js
dist/highcharts.src.js
dist/highmaps.js
dist/highmaps.src.js
dist/highstock.js
dist/highstock.src.js
```

PeterDaveHello:
Would you like to send a PR to add it?

PR #6343, opened 2015-11-29
[new][author] Add react-highcharts@5.0.6

kirjs:
Docs: https://github.com/kirjs/react-highcharts


kirjs:
See #6317

ajax/libs/react-highcharts/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "react",
+    "highcharts",
+    "graph"
+  ],
+  "readmeFilename": "README.md",
+  "gitHead": "896bf22728ae74749d3355695adb2fee63c921af",
+  "homepage": "https://github.com/kirjs/react-highcharts#readme",
+  "_id": "react-highcharts@5.0.6",
+  "_shasum": "815331d88d9f0c1b8668fcebd3fc6d9b79067bb1",
+  "_from": "react-highcharts@",
```
PeterDaveHello: Please remove line 5, 24~29, we don't need them, thanks.

kirjs: done

PeterDaveHello:
Ok, thanks, I have some questions here:

 - Don't you need `bundle` dir under `dist`?
 - Personally, I think highchart, highmaps and hightstock are different products, does it make sense to include all of them in react-highcharts? 

Thanks.

kirjs:
1. Bundles are just helpers which include Highcharts code which is already on cdnjs, so I'd avoid duplication and not include them. 
2. They historically got into the same repository, because due to complex Highcharts module structure it was easier to support them this way. I would make sense to separate them out at some point, but currently this is where the users would expect those modules to be.

PeterDaveHello:
Thanks for your explanation @kirjs

kirjs:
Thanks a lot!

Issue #5544, opened 2015-08-29
Un-recognized license(s) field by SPDX license identifier


PeterDaveHello:
Here are the libs with license(s) field but can not been recognized by SPDX license identifier([1](https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json), [2](https://spdx.org/licenses/)), we need to confirm their license(s) and try to use use a clean and clear SPDX license identifier recognized identifier in package.json.

cc #5543 @cdnjs/intern 

Library name / current license(s) field:
- [x] ScrollMagic  :  GPL-3.0+
- [x] UAParser.js  :  GPLv2
- [x] WebRupee  :  Apache License, Version 2.0
- [x] balance-text  :  Apache License, Version 2.0
- [x] bootstrap-rtl  :  Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International
- [x] chainvas  :  none
- [x] crafty  :  MIT
- [x] d3-geo-projection  :  BSD
- [x] datatables-colvis  :  MIT
- [x] datatables-fixedheader  :  MIT
- [x] datatables-tabletools  :  MIT
- [x] django.js  :  LGPL 3
- [x] dojo  :  AFLv2.1
- [x] dojo  :  BSD
- [ ] ekko-lightbox  :  MIT
- [ ] embedly-jquery  :  BSD-3-Clause
- [ ] emojione  :  (CC-BY-SA-4.0 and MIT)
- [ ] equalize.js  :  GPLv2
- [ ] evil.js  :  Public Domain
- [ ] extjs  :  GNU General Public License Version 3
- [ ] extjs  :  Open Source License Exception for Applications
- [ ] extjs  :  Open Source License Exception for Development
- [ ] froala-editor  :  CC
- [ ] froala-editor  :  Single Website
- [ ] froala-editor  :  Developer
- [ ] froala-editor  :  Business
- [ ] froala-editor  :  OEM
- [ ] geoext  :  BSD
- [x] highmaps  :  Creative Commons Attribution-NonCommercial 3.0 License.
- [x] ice  :  GPLv2
- [ ] intercom.js  :  Apache
- [x] ixjs  :  Apache License, Version 2.0
- [x] jQuery-slimScroll  :  GPL & MIT
- [x] jQuery.dotdotdot  :  MIT
- [ ] jo  :  Redistribution
- [ ] jqPlot  :  GPL
[...]

PR #6344, opened 2015-11-30
Add license of fiber #6324

Asu4ni:
source: https://github.com/linkedin/Fiber/blob/master/LICENSE.txt
#6324

### DIFF
--- a/ajax/libs/fiber/package.json
+++ b/ajax/libs/fiber/package.json
@@ -16,5 +16,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/linkedin/Fiber.git"
-  }
+  },
+  "license": "Apache License 2.0"
 }


Asu4ni:
I still don't know how to set my own branch name. It just set it patch-2, and I couldn't find option to change it. Sorry about that.

Amomo:
@Asu4ni oops, this case is a little different.

please see #5544, cdnjs use SPDX license identifier to recognize library's license.
So, it is better to use "Apache-2.0" than "Apache Software License 2".
Could you use "Apache-2.0" in the license field?

And, as cdnjs has many commits in repository, we expect you use `one commit` to finish this work.
If you can't delete the old commit, you should open a new branch to finish this work.

The branch name is fine in this time,
we found that you can't modify the branch name if you edit files directly on cdnjs' repo.
It would be fine if you edit it on your fork repo.

PR #6345, opened 2015-11-30
Add license of fiber #6324

Asu4ni:
source: https://github.com/linkedin/Fiber/blob/master/LICENSE.txt
#6324


Amomo:
Great! :+1:

PeterDaveHello:
"Add license of fiber, #6324" will be better

PR #6346, opened 2015-11-30
add missing license of angular-route-segment

pvnr0082t:
PR for #6324
license info ref: https://github.com/artch/angular-route-segment#license


Amomo:
Good job! :+1:

PR #6355, opened 2015-12-01
add missing license of architect

pvnr0082t:
PR for #6324
license info ref: https://github.com/EtienneLem/architect/blob/master/MIT-LICENSE


Amomo:
Good try! Better than your first PR. :smile: 
Nice to see a well named branch name and good reference!

PR #6369, opened 2015-12-02
[new] [author] Added chess.js v0.9.3

jhlywa:
Some of the chess.js users have requested that it be added to cdnjs (jhlywa/chess.js#73)


PeterDaveHello:
Hi @jhlywa, thanks for your PR, I wonder if we need the test.js file here?

jhlywa:
The tests were automatically included by cdnjs-importer.  I have removed them.

PeterDaveHello:
Okay, would you please squash/fixup them into only one commit? Thanks.

jhlywa:
Ok, I switched to npm instead of git autoupdate, fixed whitespace issues in package.json, and squashed.

PR #6388, opened 2015-12-04
Added gridly@1.1.0

IonicaBizau:
Added [`gridly`](https://github.com/IonicaBizau/gridly)––the 157B grid system. :dizzy: 
Fixes https://github.com/IonicaBizau/gridly/issues/1.

### DIFF
--- a/ajax/libs/gridly/package.json
+++ b/ajax/libs/gridly/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "gridly",
+  "version": "1.1.0",
+  "description": "The minimal (~157 bytes) grid system for modern browsers.",
+  "main": "lib/gridly.css",
+  "directories": {
+    "example": "example"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/IonicaBizau/gridly.git"
+  },
+  "keywords": [
+    "gridly",
+    "grid",
+    "css",
+    "modern",
+    "browsers"
+  ],
+  "author": "Ionică Bizău <bizauionica@gmail.com> (http://ionicabizau.net)",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/IonicaBizau/gridly/issues"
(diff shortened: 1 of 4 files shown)


ajax/libs/gridly/package.json
```diff
@@ -0,0 +1,132 @@
...
+    "modern",
+    "browsers"
+  ],
+  "author": "Ionică Bizău <bizauionica@gmail.com> (http://ionicabizau.net)",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/IonicaBizau/gridly/issues"
+  },
+  "homepage": "https://github.com/IonicaBizau/gridly#readme",
+  "blah": {
```
PeterDaveHello: We should remove this `blah` which has no meaning for cdnjs.

ajax/libs/gridly/package.json
```diff
@@ -0,0 +1,132 @@
...
+  "npmName": "gridly",
+  "npmFileMap": [
+    {
+      "basePath": "/dist",
+      "files": [
+        "**/*"
+      ]
+    }
+  ],
+  "filename": "/gridly-col-widths.min.css"
```
PeterDaveHello: There is no need to use a leading slash for the files under the root.

IonicaBizau: @PeterDaveHello Recommitted. :smile:

ajax/libs/gridly/package.json
IonicaBizau: Haha, hoped you will not notice that! I guess this should be part of the unit testing, shouldn't it? :joy:

IonicaBizau:
:tada: :rocket: Thanks!

PeterDaveHello:
Now you can access it via:
https://cdnjs.com/libraries/gridly
https://cdnjs.cloudflare.com/ajax/libs/gridly/1.1.0/gridly-col-widths.min.css

PR #6433, opened 2015-12-08
update autoupdate config of summernote

outsideris:
summernote wants to deploy language files alongside dist files. summernote/summernote#1449

### DIFF
--- a/ajax/libs/summernote/package.json
+++ b/ajax/libs/summernote/package.json
@@ -13,9 +13,9 @@
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/summernote/summernote.git",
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
@outsideris thanks for your PR however, the changes of the directory/file structure will cause breaking changes between different versions, is it possible to not to break it but just only add more files under dist? Thanks.

outsideris:
@PeterDaveHello I got what you means. I was intended to make same directory structure between bower, npm and cdnjs, but you are right. I will put all files in `dist` directory in next release. So, I close this PR. Thank you.

PeterDaveHello:
Thanks you @outsideris

PR #6435, opened 2015-12-08
qTip2 - Update package.json - npmName + npmFileMap + homepage

Craga89:
Add auto-updating `npmName` and `npmFileMap` entries

PR #6436, opened 2015-12-08
qTip2 - Update package.json - npmName + npmFileMap

Craga89:
Add auto-updating `npmName` and `npmFileMap` entries


PeterDaveHello:
As we disscussed in #6435, the lib's name should be mentioned in the first line of your commit message, and the homepage is just out-dated, which will be redirected to the new one, I don't think it's "wrong".

Craga89:
Gah sorry, thought you meant the actual pull request. Will re-request

PR #6447, opened 2015-12-09
Added: Dummyjs

spyshiv:
**Library name: Dummyjs** 
**Git repository url: https://github.com/spyshiv/dummyjs**
**npm package url(optional):** 
**License(s):GNU GENERAL PUBLIC LICENSE Version 2, June 1991**
**Official homepage:http://spyshiv.github.io/dummyjs/**
**Wanna say something? Leave message here:**Dummyjs is js library to generate dummy words, dummy sentences and dummy paragraphs which are very useful while writing a webpage and instead of content you have to write some dummy text like...lorem ipsum dolor sit etc..using dummyjs can save the much time of web developer


spyshiv:
Something went wrong

Issue #6350, opened 2015-11-30
[Request] JsRender


abdulhannanali:
**JsRender** 
**https://github.com/BorisMoore/jsrender**
**npm package url(optional):** 
**MIT**
**http://www.jsviews.com/**
**Please upload it and I love CdnJS**

=====================
Notes from cdnjs maintainer:
You are welcome to add a library via sending pull request,
it'll be faster then just opening a request issue,
and please don't forget to read the guidelines for contributing, thanks!!

mably:
+1

abdulhannanali:
thanks everyone

abdulhannanali:
@maruilian11 special thanks to you

maruilian11:
@abdulhannanali :+1:

BorisMoore:
I have just updated JsRender to v0.9.71. It should fix the autoupdate issue. (See https://github.com/BorisMoore/jsrender/issues/278#issuecomment-162769636). Thanks for your work on cdnjs!

PeterDaveHello:
https://github.com/cdnjs/cdnjs/pull/6499#issuecomment-166228870

PR #6499, opened 2015-12-18
add jsrender v1.0.0-rc.70

maruilian11:
**git repo url:** https://github.com/BorisMoore/jsrender
**Watch 150
  Star 1,720
 Fork 285**
@Amomo  could u help me check this PR for #6350 ? thank you!
this lib do have npm package, but when i use npm install jsrender, it would be https://github.com/defunctzombie/node-jsrender .
the author found the problem recently, and they may fix it in next update.
i will ask the author if they can tag us when they fix it 
then, we can change the config to npm auto-update :smiley:
- [x] Not found on cdnjs repo
- [x] No already exist issue and PR
- [x] Notable popularity
- [x] Project has public repository on famous online hosting platform (or been hosted on npm) 
- [x] Has valid tags for each versions (for git auto-update)
- [x] Auto-update setup
- [x] Auto-update target/source is valid.
- [x] Auto-update filemap is correct


Amomo:
@PeterDaveHello I think this PR is fine, we can change to npm auto-update (if we need) when next version released. Could you take a look at it? Thank you! :-)

PeterDaveHello:
> this lib do have npm package, but when i use npm install jsrender, it would be https://github.com/defunctzombie/node-jsrender .
> the author found the problem recently, and they may fix it in next update.
> i will ask the author if they can tag us when they fix it 

Why don't  you just give us the reference? I don't think you said it clare.

maruilian11:
@PeterDaveHello @Amomo reference here
https://github.com/BorisMoore/jsrender/issues/278#issuecomment-162769636

PeterDaveHello:
1. Your auto-update config includes jsrender-node.js, which is no need to be hosted on cdn. Should use .gitignore to ignore it or use another config. See 9cfaec2e1a4f3b289920067ffa489fe4072b2076.
2. You should add and use 0.9.71 in package.json not a rc version, unless there is no stable version. See 4423fbd43f1fafc067e644800bfa18e511253dc2.

maruilian11:
ok. i see, thanks for reminding.

PR #6517, opened 2015-12-18
Added SystemJS 0.19.9

KarlDirck:
From dist directory of https://github.com/systemjs/systemjs/releases/tag/0.19.9
The associated package.json has been updated with the source, version, dependencies, etc.

PR #6532, opened 2015-12-21
fix npm auto-update config typo of systemjs

maruilian11:
because #6517, i found there is a typo error, so the latest version can't be updated.

### DIFF
--- a/ajax/libs/systemjs/package.json
+++ b/ajax/libs/systemjs/package.json
@@ -15,9 +15,9 @@
   "npmName": "systemjs",
   "npmFileMap": [
     {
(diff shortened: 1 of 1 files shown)


ajax/libs/systemjs/package.json
```diff
@@ -15,9 +15,9 @@
   "npmName": "systemjs",
   "npmFileMap": [
     {
-      "basePath": "/dist",
+      "basePath": "dist",
```
PeterDaveHello: This is not a typo, but `basePath` without prefix and postfix slash is good.

PR #6675, opened 2016-01-08
[author][add] nice-validator@0.10.0

niceue:
The repo: https://github.com/niceue/nice-validator


PeterDaveHello:
Oops, should remove the second Merge commit.

PeterDaveHello:
The first commit in this PR, it should be based on our latest master branch, but seems it has no parent ...

PR #6676, opened 2016-01-09
Enrich ICanHaz.js' package.json(field: license)

Kakerucchi:
##


PeterDaveHello:
hmmm ... strange, why did you use @LeaYeh's identify to do this commit?

LeaYeh:
how did you get that XD?
that commit still has something wrong and i didnt fix that.

PeterDaveHello:
Close due to no response and as we already added the license

PR #6733, opened 2016-01-18
[author][new] Add babel-standalone

Daniel15:
Adds [babel-standalone](https://github.com/Daniel15/babel-standalone), which is a replacement for babel-browser (which itself was a replacement for babel-core usage in the browser).

Auto-update from npm: https://www.npmjs.com/package/babel-standalone. Files from npm are the same as in the release section of Github:
 - https://github.com/Daniel15/babel-standalone/releases/download/release-6.4.4/babel.js
 - https://github.com/Daniel15/babel-standalone/releases/download/release-6.4.4/babel.min.js


PeterDaveHello:
Thanks @Daniel15, would you consider to add some keywords in its package.json?

Daniel15:
Good catch, I deleted the keywords while removing unnecessary stuff from the package.json. I should add them back in :)

PeterDaveHello:
@Daniel15 :+1:

PR #6748, opened 2016-01-19
[new] add typicons

equinox7:
http://www.typicons.com/
https://github.com/stephenhutchings/typicons.font


equinox7:
Created new pull request with one commit.

PeterDaveHello:
Don't we need svg and png images under src? Thanks.

equinox7:
@PeterDaveHello Personally I do not need src images. This is not the only icon font without src images in it. For example **octicons** has only fonts and styles in it too.

PeterDaveHello:
@equinox7 I just want to confirm if the png and svg files should also be distributed, maybe we can confirm with its author?

equinox7:
@PeterDaveHello There are [usage notes](http://www.typicons.com/more/how-to-use-typicons/) only for font icons. If you want, I can try contact author.

PeterDaveHello:
@equinox7 okay I will take a look, thanks a lot

PeterDaveHello:
LGTM, let's give it a try.

PR #6765, opened 2016-01-20
Add missing "filename" requirement for autoupdate

Mottie:
While performing an `npm test`, I discovered that a `"filename"` property is required for an autoupdate setup:

``` shell
    my-repo
      ✗ my-repo package.json is valid
        » my-repo is not a valid cdnjs package.json format:
          » NPM package.json
                Property is required (true): /properties/filename // valid-packages-test.js:82

      ✗ my-repo: filename from package.json exists
        » ./ajax/libs/my-repo/1.0.0/undefined does not exist but is referenced in package.json! // valid-packages-test.js:101
```


PeterDaveHello:
@Mottie all the package.json on cdnjs must have their own filename info, not only the auto-update libs.

Mottie:
Resubmitted as PR #9842

PR #6804, opened 2016-01-25
point cookiejar.js to cookiejar.min.js, #5195

LeaYeh:
Hi @Amomo  , From #5195, I add the minify files to each version of cookiejar, and point cookiejar.js to cookiejar.min.js in package.json.


PeterDaveHello:
What's the reason why to touch ajax/libs/cookiejar/0.5/cookiejar.min.js?

LeaYeh:
do not need to make same structure in each version?

Amomo:
Hi @LeaYeh , sorry for late.
I think you need to separate your commit into 2 parts.
One is for adding minified version for v0.5 & v0.5.1, and another is for modify the "filename" field in package.json.

PeterDaveHello:
> do not need to make same structure in each version?

What's the relation between making same structure in each version and touching unrelated file?

PR #6922, opened 2016-02-04
update angular-md5 auto-update

wil93:


PeterDaveHello:
Hi @wil93, thanks for your PR, we prefer to use npm auto-update, and I just found that the `npmName` field is missing, would you please just help us add the missing info in package? Thanks a lot.

wil93:
OK, done. I force-pushed the patch to add just the `npmName` field (also reverting the update from 0.1.7 to 0.1.10). I followed this guidelines: https://github.com/cdnjs/cdnjs/blob/master/documents/autoupdate.md

PeterDaveHello:
Thank you @wil93

PR #6994, opened 2016-02-16
[author][new] Added ConstraintJS Library@0.9.6

soney:
Origin of files: http://cjs.from.so/


ajax/libs/constraintjs/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "constraintjs",
+  "npmName": "constraintjs",
+  "npmFileMap": [
+    {
+      "basePath": "/dist/",
+      "files": [
+        "jquery.*"
```
PeterDaveHello: I didn't see you commit `jquery.*`, maybe the config is wrong?

soney: Thanks; removing

PeterDaveHello:
Would you please try to enable auto-update on it? https://github.com/cdnjs/cdnjs/blob/master/documents/autoupdate.md

Maybe `repositories` and `authors` can be object not array, there is only one object there.

PeterDaveHello:
Any updates here? Thanks!

PR #7188, opened 2016-03-07
Change mathjs library to npm update script

josdejong:
I saw cdnjs hadn't picked up the latest release of mathjs, v3.0.0. 

This PR changes package.json to the preferred npm update script and updates to v3.0.0.

Thanks!


josdejong:
Hm I could have known that this would fail, I updated the version number too without adding the new files for v3. I will create a new PR and only update the npmFileMap part.

PR #7220, opened 2016-03-09
Correct Select2 home page

kevin-brown:
The home page for Select2 was switched during the 4.0.0 release to the new location at https://select2.github.io. This finally fixes the homepage link in the `package.json` here, since cdnjs never updated it after it was changed in the NPM package.


PeterDaveHello:
Thanks @kevin-brown

Issue #7222, opened 2016-03-09
[Request] jQuery Datetimepicker Full


evromalarkey:
**Library name:** jquery-datetimepicker
**Git repository url:** https://github.com/xdan/datetimepicker
**npm package url(optional):** https://github.com/xdan/datetimepicker/blob/master/package.json
**License(s):**
**Official homepage:**
**Wanna say something? Leave message here:**
The library is already on cdnjs (https://cdnjs.com/libraries/jquery-datetimepicker), but that's no longer the full version. -  jquery.datetimepicker.full.min.js is missing there

=====================
Notes from cdnjs maintainer:
Please read the README.md and CONTRIBUTING.md document first.

You are welcome to add a library via sending pull request,
it'll be faster then just opening a request issue,
and please don't forget to read the guidelines for contributing, thanks!!

evromalarkey:
unfortunately I still can't see the full version on cdnjs
http://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.1/jquery.datetimepicker.js still lacks the DateFormatter dependency, so it's not working - here is how it should be https://github.com/xdan/datetimepicker/blob/master/build/jquery.datetimepicker.full.js

PeterDaveHello:
@evromalarkey I will take a look at it. Thanks for report.

PeterDaveHello:
@evromalarkey have you tried this?
https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.3/build/jquery.datetimepicker.full.min.js

PR #7345, opened 2016-03-22
update auto-update config of jquery-datetimepicker

maruilian11:
@Amomo could you help me check this PR for #7222 ? thank you!
we already hosted this lib, just need to add full files.


PeterDaveHello:
add full.js v2.5.1 ? Actually you don't need to have commit ae858c2, I'll handle it, and seems you only added the missing files in v2.5.1 version, and missed the old versions.

maruilian11:
@PeterDaveHello updated.

Issue #7263, opened 2016-03-12
Transfer angular-chart.js to npm auto-update


PeterDaveHello:
https://github.com/jtblin/angular-chart.js - npm package: angular-chart.js

Should also try to add cdnjs here:
https://github.com/jtblin/angular-chart.js/tree/1.0.0-alpha1#cdn

although we don't support the `latest` version syntax,

PR #7365, opened 2016-03-25
modify auto-update config of angular-chart.js from git to npm

maruilian11:
for #7263

### DIFF
--- a/ajax/libs/angular-chart.js/package.json
+++ b/ajax/libs/angular-chart.js/package.json
@@ -9,12 +9,13 @@
     "url": "git://github.com/jtblin/angular-chart.js.git"
   },
   "license": "BSD",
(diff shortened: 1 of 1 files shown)


ajax/libs/angular-chart.js/package.json
```diff
@@ -9,12 +9,13 @@
...
-    "files": [
-      "**/*.+(css|js|map)"
-    ]
-  }
+  "npmName": "angular-chart.js",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "**/*.+(css|js|map)"
```
PeterDaveHello: Try to drop `*.gz` only.

maruilian11:
@PeterDaveHello  update.

PR #7979, opened 2016-05-19
Update wow's license

graingert:


ajax/libs/wow/package.json
```diff
@@ -22,5 +22,5 @@
       "*"
     ]
   },
-  "license": "MIT"
+  "license": "GPLv3"
```
PeterDaveHello: Please use "GPL-3.0"

PR #8090, opened 2016-05-30
Update file structure of UPB package

DaAwesomeP:
I am the author of this library. In the most recent version I moved from a `dist` folder to `lib` folder. There are no dependencies to compile, so it works fine. This pull fixes the auto-update to reflect the structure change.

### DIFF
--- a/ajax/libs/upb/package.json
+++ b/ajax/libs/upb/package.json
@@ -19,20 +19,13 @@
   "npmName": "upb",
   "npmFileMap": [
     {
-      "basePath": "/",
+      "basePath": "lib",
       "files": [
-        "upb.js"
-      ]
-    },
-    {
-      "basePath": "/dist/",
-      "files": [
-        ".js",
-        ".map"
+        "index.js"
       ]
     }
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
So we only need to grab `index.js`? Is it possible not to change the filename? So that users can easily change the version by the version number in url, thanks.

DaAwesomeP:
This is the standard/preferred file structure for Node/NPM libraries, which this works with as well as with the browser. I've seen it in lots of Bower repos too. Does not listing the file in the `files` attribute and then changing the `filename` attribute in the cdnjs package.json take care of this?

PeterDaveHello:
Nope we don't change any filenames in autoupdate process, if the old ones work good, can we just keep that? So the users can easily upgrade or downgrade without breaking the sites since the filename should also be updated not only the version.

DaAwesomeP:
I guess I can add back a `dist` folder.

PeterDaveHello:
Thanks @DaAwesomeP

PR #8133, opened 2016-06-03
Update package.json for jquery.lazy

eisbehr-:
The Library is already available on cdnjs, I'll just added the new plugin file to cdn too.

### DIFF
--- a/ajax/libs/jquery.lazy/package.json
+++ b/ajax/libs/jquery.lazy/package.json
@@ -45,7 +45,9 @@
     "basePath": "",
     "files": [
       "jquery.lazy.min.js",
-      "jquery.lazy.js"
+      "jquery.lazy.js",
+      "jquery.lazy.plugins.min.js",
+      "jquery.lazy.plugins.js"
     ]
   }
 }


PeterDaveHello:
@eisbehr- what about using this single rule: `"jquery*.js",`?

eisbehr-:
Yes, this would be a solution too! :)

Is it even possible, and allowed by cdnjs, to add `"plugins/jquery*.js"` too?
With this even all plugins would be hosted as single files.

PeterDaveHello:
@eisbehr- no problem, but just please don't add the famous "jQuery" in it :smile:

PeterDaveHello:
Another method is to create `dist` folder for distribution purpose, and we can use pattern `"**/*"` to grab all the files in it.

eisbehr-:
Okay, thanks. I'll create a new PR tomorrow adding those and change it to mask usage. 

A dist folder is no option for now, because I don't want to change the stucture in a working repo. ;)

PeterDaveHello:
@eisbehr- Okay, thanks.

PR #8147, opened 2016-06-04
[author] Update package.json for jquery.lazy

eisbehr-:
Add plugin files and update dependencies.
Use file rules as suggested in #8133.


eisbehr-:
Any problems here? :smiley:
I would like to push an update to git these days, so an updated package would be great. Thank you!

PR #8269, opened 2016-06-17
Add missing anythingslider theme images

gttds:
Image referenced by theme-minimalist-round.css and theme-minimalist-square.css.
Source https://github.com/CSS-Tricks/AnythingSlider/blob/master/images/arrows-minimalist.png


PeterDaveHello:
Would you please also update the auto-update config in package.json to make sure we won't miss this image in future?

gttds:
Yes no problem. Have now updated package.json as well.

PeterDaveHello:
@gttds Thanks.

PeterDaveHello:
@gttds do you know that if we may also need other images under `images` folder? Thanks.

gttds:
@PeterDaveHello yes, two other images. arrows-metallic.png and cs-portfolio.png which are referenced by theme-metallic.css and theme-cs-portfolio.css respectively.

PeterDaveHello:
So can you also add them? And maybe change the rule of images to single `"images/*.png"`, thanks.

gttds:
No problem. Have done that and also added a gif image which is needed by one of the themes.

PeterDaveHello:
Thanks! What about `default.png`?

gttds:
Shouldn't be an issue as the updated rule `"images/*.+(gif|png)"` ensures that `default.png` will continue to be included in the package.

PeterDaveHello:
@gttds but it's not in the commit, auto-update config will not affact the already exist versions.

gttds:
It's already in the existing versions :) https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/anythingslider/1.9.6/images/default.png

PeterDaveHello:
@gttds Oh okay, my mistake, thanks.

PR #8402, opened 2016-06-30
Selectize.js: Update ref to moved repo

joallard:
Not really sure how to proceed here, but I hope making this change will allow the lib to auto-update. I'm a maintainer of Selectize at https://github.com/selectize/selectize.js/.


PeterDaveHello:
Thanks @joallard :100:

Issue #5422, opened 2015-08-19
Lack of keywords in lib's package.json


PeterDaveHello:
There is no `keywords` field in these libs, maybe we can add keywords for them, and **also send feedback to upstream**.

 - [ ] angular-cache
 - [ ] angular-hal
 - [ ] angular-ui-calendar
 - [ ] angular-ui-select
 - [ ] angular-ui-sortable
 - [ ] angular-utf8-base64
 - [ ] antimoderate
 - [ ] Base64
 - [ ] booking-js
 - [ ] bootbox.js
 - [ ] bootstrap-daterangepicker
 - [ ] chaplin
 - [ ] circliful
 - [ ] clappr
 - [ ] depot
 - [ ] device.js
 - [ ] django.js
 - [ ] documentup
 - [ ] dragdealer
 - [ ] dragula
 - [ ] engine.io-client
 - [ ] eve.js
 - [ ] eventable
 - [ ] expect.js
 - [ ] fluidvids.js
 - [ ] formstone
 - [ ] fpsmeter
 - [ ] github-repo-widget
 - [ ] graphiql
 - [ ] grommet-index
 - [ ] gulp
 - [ ] gumshoe
 - [ ] hls.js
 - [ ] instantsearch.js
 - [ ] interpolate.js
 - [ ] jmpress
 - [ ] joyride
 - [ ] jquery-fracs
 - [ ] jquery-mobile-datebox
 - [ ] jquery-overscroll
 - [ ] jquery-scrollpanel
 - [ ] jquery-twinkle
 - [ ] jquery.photocols
 - [ ] jquery.turbolinks
 - [ ] js-data-angular
 - [ ] jstat
 - [ ] knockout-postbox
 - [ ] label.css
 - [ ] Ladda
 - [ ] legojs
 - [ ] linkurious.js
 - [ ] localStorage
 - [ ] lrsjng.jquery-qrcode
 - [ ] midi.js
 - [ ] midnight.js
 - [ ] ng-tasty
 - [ ] oidc-client
 - [ ] outdated-browser
 - [ ] papier
 - [ ] r2d3
 - [ ] ramjet.js
 - [ ] range.js
 - [ ] repo.js
 - [ ] requirejs-plugins
 - [ ] sbt
 - [ ] scaleapp
 - [ ] shaka-player
 - [ ] sigma.js
 - [ ] slidebars
 - [ ] snabbt.js
 - [ ] snackbarjs
 - [ ] social-sharing
[...]

PR #8469, opened 2016-07-07
Modify keywords field of attrchange, cc #5422

x09326:
PR for #5422 
@pvnr0082t please help me review it, thanks.


ajax/libs/attrchange/package.json
```diff
@@ -11,6 +11,12 @@
     "onchange",
     "onpropertychange"
   ],
+  "keywords": [
```
pvnr0082t: @x09326 I think you can reference the upstream.
Line 6-12 of https://github.com/meetselva/attrchange/blob/master/attrchange.jquery.json

PR #8483, opened 2016-07-09
Add keywords of backbone.modal, cc #5422

x09326:
PR for #5422 
@kennynaoh please help me review it, thanks.


ajax/libs/backbone.modal/package.json
```diff
@@ -9,6 +9,11 @@
     "type": "git",
     "url": "git://github.com/awkward/backbone.modal.git"
   },
+  "keywords": [
+    "boilerplate",
```
PeterDaveHello: `modal` will be better than `boilerplate`

PR #8841, opened 2016-08-26
manually add babel-core old version (5.8.35 / 5.8.38)

iwdmb:
I recently wanted to use babel-core latest version(5.x).

But, I can not find babel-core 5.8.38.

[v5.8.35](https://github.com/babel/babel/commit/0dcb25b79a1500d1670550be5521ce31638344e1)
[v5.8.38](https://github.com/babel/babel/commit/3ef6e59cce73c600580d5035780f27d68b16c1e1)


PeterDaveHello:
That's because we don't use Babel-core as the source from 99c5a15b85ee94a3b76969299e11753da20fe96c, I don't think manually add these two versions is a good idea.

PeterDaveHello:
Looks good to merge :+1:

Issue #4599, opened 2015-04-30
Support SRI (Subresource Integrity)


nikolay:
You CloudFlare just had [an article](https://blog.cloudflare.com/an-introduction-to-javascript-based-ddos/) regarding SRI, so, you should start publishing the `integrity` value for each.

PeterDaveHello:
cc @terinjokes

fletom:
+1

jonathanKingston:
So is the script which imports the resource data to be indexed by algolia on GitHub? (I can't see it anywhere).

If it is then I can help with making the relevant pull requests as have done other work which does this.

Does cdnjs ever change the libraries like Google has done in the past?

PeterDaveHello:
@jonathanKingston 
> Does cdnjs ever change the libraries like Google has done in the past?

What's the change actually? Thanks.

jonathanKingston:
@PeterDaveHello it patches certain versions of their code if there are critical versions (Don't quote me on this but I think that was even if you provided a full semver version path). Either way they offer the ability to say Jquery@2 and it will self update.

I assume cdnjs doesn't do this ever right?

PeterDaveHello:
Yeah, we don't.

jonathanKingston:
@PeterDaveHello how does the search get populated as I think that would be the best candidate for storing the SRI hashes precalculated so when the client loads the page they can see the files with the hashes.

PeterDaveHello:
cc @thomasdavis @drewfreyling @ryankirkman for comments

jonathanKingston:
How does public/packages.min.json get populated? I can see the indexing code mostly uses that with the json containing file size etc.

xPaw:
Chrome 45+ now supports SRI, https://googlechrome.github.io/samples/subresource-integrity/index.html

PeterDaveHello:
@terinjokes do you have any idea? Thanks!

mozfreddyb:
What's blocking this? Would love to use cdnjs with subresource integrity ;D

terinjokes:
I'm hoping to chat with @PeterDaveHello soon to see where I can help in getting SRI displayed on cdnjs.com.

jackryu:
Hi Team,Please share C# code - for bind chart with SQL database. In asp.net.

Regards,
Abhishek Tomar




On Tue, Sep 22, 2015 at 1:08 PM -0700, "Terin Stock" <notifications@github.com> wrote:
[...]

PR #8874, opened 2016-08-30
impliment SRI (again) for js and css with sha256

PeterDaveHello:
for #4599

Now write SRI data by "library name" / "library version" and store in the `sri` directory under the `new-website` repo.

### DIFF
--- a/build/packages.json.js
+++ b/build/packages.json.js
@@ -18,6 +18,7 @@ var feed = new RSS({
     author: 'cdnjs team'
 });
 var exec=require('child_process').exec;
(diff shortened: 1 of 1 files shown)


drewfreyling:
@PeterDaveHello would this have been easier to use a NPM package rather than rely on openssl for the hash generation? Something like https://www.npmjs.com/package/node-sri

PeterDaveHello:
@drewfreyling as I said I found 2 npm packages got different hash on certain files, using execSync is also easy and reliable.

PR #8899, opened 2016-09-02
Update Visibility.js to 1.2.4

ai:
@PeterDaveHello new version of Visibility.js (I am a author).

Wait for this PR before release.


PeterDaveHello:
@ai was the main filename changed?

ai:
Sorry. I updated PR.

PeterDaveHello:
@ai is it possible to setup the auto-update config here? So no more manaual updates needed.

ai:
@PeterDaveHello I didn’t really develop this library, only support it. So I think we will have 1-3 releases in next few year. In this case, I think, auto-update config will take more time :)

PeterDaveHello:
Sorry, I didn't get it, what do you mean "will take more time" here? Actually we hope all the libs have auto-update config which can decrease the workload of our maintainers, and you don't need to submit new PR, waiting for revuiew, the new versions will be available within just 30 mins after released, should be better.

ai:
Could you accept this release first? It block version release ;).

ai:
I don’t want to add auto-update config, because I think it will be the last Visibility.js release ;)

ai:
Thanks =^_^=

PR #8908, opened 2016-09-04
Updated mark.js v8.1.0

julmot:
I've forgot to increase the minor version before creating the tag.


PeterDaveHello:
@julmot did you just re-tagged v8.1.0 in the repo?

PeterDaveHello:
Sorry, I have to revert this commit, since we started to use SRI for security enhancement, this change can just break the current users of mark.js v8.1.0 on cdnjs.

julmot:
@PeterDaveHello You're right!

PR #5201, opened 2015-08-01
[author] stampit: cleaned up and updated auto-update information

sethlivingston:
- Removed non-production releases
- Switched repository from stampit-org/stampit (CommonJS) to stampit-org/bower (UMD)
- Updated auto-update information to use git instead of npm

PR #9076, opened 2016-09-23
[author] stampit: update autoupdate and version to v3

FredyC:
There was a PR #5201 last year which pointed stampit library to no longer maintained repository (https://github.com/stampit-org/stampit-bower). Today we have released new major version and want to switch to the official repository (https://github.com/stampit-org/stampit).

We don't have distribution file in repo, so I haven't included `autoupdate` key in package.json. Is it ok like that? Will auto update work just based on this?


PeterDaveHello:
@FredyC Thanks for your PR! @cdnjs/intern2 please help reviewer this PR, thanks.

FredyC:
Just noticed I've probably forgot to add `filename` key there. I would like to amend and push force to that commit to keep it small, but how to do that easily? Do I need to clone whole repo which has GB of data?

PeterDaveHello:
@FredyC You can just modify it online, no need to clone the repo, GitHub now allows maintainer to squash commits in PR, thanks.

ajax/libs/stampit/package.json
```diff
@@ -1,39 +1,37 @@
...
     "object oriented",
     "browser",
     "inheritance",
-    "oo",
+    "oop",
     "node",
     "factory",
-    "class"
+    "class",
+    "stamp"
```
abhisekp: I suggest to add some more keywords like `functional`, `compose`, `constructor`

FredyC: Well these are keyword from official repo, so perhaps first add it there and then we can go with it here :)

FredyC:
I am bit confused about the circleci fail...

> stampit: filename from package.json exists 
>        » ./ajax/libs/stampit/3.0.0/stampit.umd.min.js does not exist but is referenced in package.json! // valid-packages-test.js:105 

Does that mean I need to copy whole release into this PR? I thought that's job of autoupdate feature. Or shall I just remove version from package.json?

pvnr0082t:
@FredyC You can just remove version info from `package.json`.

ajax/libs/stampit/package.json
```diff
@@ -1,39 +1,37 @@
...
-      "stampit.min.js"
-    ]
+    "url": "git@github.com:stampit-org/stampit.git"
   },
-  "filename": "stampit.min.js"
+  "filename": "stampit.umd.min.js",
+  "npmName": "stampit",
+  "npmFileMap": [
+    {
+      "basePath": "/dist/",
```
pvnr0082t: Let's use `"dist"` here.

PeterDaveHello:
@FredyC Yes, you need to add the whole release files to the PR, or, you can just remove the version info in package.json so that the test will ignore this part, thanks.

ajax/libs/stampit/package.json
```diff
@@ -1,39 +1,37 @@
...
+      "files": [
+        "stampit.umd.js",
+        "stampit.umd.min.js"
```
pvnr0082t: @FredyC If the files under `dist` folder need to be added in cdnjs, let's use `"**/*"`. 
If you just want to add `stampit.umd.js` and `stampit.umd.min.js`, I suggest to use `"stampit.umd*"`. It will be more flexible. Thanks.

ajax/libs/stampit/package.json
FredyC: I see. Removed, hopefully now it will be correct finally :)

ajax/libs/stampit/package.json
FredyC: Hmm, there are other files in dist for sure. I don't see much of difference if there will be one glob line or two specific ones :)

FredyC: Ok, circleci is finally passing, so unless you insist on this change, I guess it's good to go.

koresar:
Thanks @FredyC for the hard work. This is priceless!

pvnr0082t:
LGTM

PeterDaveHello:
@FredyC sorry that I said something wrong here, removal of version will only work for the first the a lib been added, so I'll need you add it back, if you don't like to add the files in the PR, please just leave filename field as well, I'll update it when import the new versions, thanks.

ajax/libs/stampit/package.json
PeterDaveHello: Should be `stampit.umd*` here, stampit.umd.js mentioned the map file, some browsers may got error if map file is not found.

FredyC:
Ok, so when I return version number and add dist files into PR, it will then work correctly for future releases? In that case I can do it.

PeterDaveHello:
@FredyC correct, thanks

FredyC:
@PeterDaveHello Hm, I've just realized that I cannot probably add new file to the branch through GH interface, can I? In that case, would it be easier for you since you probably have repo cloned?

PeterDaveHello:
@FredyC of course you can, trust me, GitHub allow you to upload file directly, just make sure they are in the same branch.

FredyC:
Alright, it's green with added releases files and version, ready to merge I would say. Thanks @PeterDaveHello and @pvnr0082t for assistance.

PeterDaveHello:
@FredyC Thanks, we will review it again ASAP, thanks for your hard working.

pvnr0082t:
LGTM

ajax/libs/stampit/package.json
```diff
@@ -1,39 +1,36 @@
...
   ],
   "repository": {
     "type": "git",
-    "url": "git://github.com/stampit-org/stampit-bower.git"
-  },
-  "dependencies": {
-    "lodash": "^3.9.1",
-    "supermixer": "^1.0.2"
-  },
-  "license": "MIT",
```
PeterDaveHello: Any reason to remove the license info?

FredyC:
@PeterDaveHello license is not removed, just moved to the top ;)

PeterDaveHello:
@FredyC hmmm ... okay, I just don't want to touch unrelated info to make the commit history harder to be traced.

koresar:
@PeterDaveHello Sorry about a misplaced License. Looks like I did it about a year ago.

Super timely @FredyC ! Thanks for the great leading. 👍 1

PR #9080, opened 2016-09-23
transfer jquery.tablesorter to npm auto-update

PeterDaveHello:
npm auto-update is faster, however just found that jquery.tablesorter uses different file format between git repo and npm package, in git repo, the dist files use unix format, but in npm pacakge, the files use windows/dos format, which is strange, no matter you use sha256 or md5 to hash the files, you'll get different result from two different sources, so we need this problem been solved first, then we can transfer this lib to npm auto-update.

cc @cdnjs/intern2

### DIFF
--- a/ajax/libs/jquery.tablesorter/package.json
+++ b/ajax/libs/jquery.tablesorter/package.json
@@ -31,12 +31,13 @@
     "type": "git",
     "url": "https://github.com/Mottie/tablesorter.git"
   },
-  "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/Mottie/tablesorter.git",
-    "basePath": "dist",
-    "files": [
-      "**/*"
-    ]
-  }
+  "npmName": "tablesorter",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "**/!(*.less)"
+      ]
+    }
+  ]
 }


PeterDaveHello:
@Mottie is it possible to use unix file format in the npm package for the future releases? At least make them the same in both git repo and npm package, it'll make sense, and as we started to serve SRI feature for a while, should better make sure they are 100% same, thank you!

Mottie:
Hi!

I have no idea why npm would use a windows/dos format. It was not intentional. Do you know how to switch formats?

Anyway, I've switched over to using Atom, so I am trying to make sure all my files are properly formatted as Unix and UTF-8. My old editor would always set the document to ANSI 1252 and DOS format after I loaded it, and supposedly convert the files back to their original format when saved.

PeterDaveHello:
@Mottie I don't think npm did that, I didn't find the same problem on the other libs yet. The format could be in your editor setting. Thanks.

PeterDaveHello:
Hello @Mottie, I saw the new release, but the problem still there, is there any update of this issue? Many thanks.

Mottie:
Yeah, I asked someone more familiar with the internal npm workings. The packed file that npm is creating contains CRLF feeds. I'm not sure why my grunt build process is adding them.

I'll try to fix this over the weekend. If I can't I won't be able to get to it for a month because I'm going on vacation overseas with only my nook 😁.

PeterDaveHello:
@Mottie Thanks! I got it!

PeterDaveHello:
Any updates here? Thanks!

Mottie:
There are only 3 unminified files in the `dist` folder and I'm not seeing any CRLFs in those files... is this still an issue?

PeterDaveHello:
Maybe we can just leave it clone.

PR #9205, opened 2016-10-07
Updated the repository url to the new organization

acrobat:
The repository was moved from eternicode to uxsolutions. Github does redirect the old repository url, but to cleanup this config and avoid useless redirects I've updated the repository urls


PeterDaveHello:
@acrobat Thanks.

PR #9207, opened 2016-10-07
Update autoupdate basePath in annyang

TalAter:
Updating existing library after directory structure change.

Files used to be in root directory, now they are in [`dist/`](https://github.com/TalAter/annyang/tree/master/dist)

### DIFF
--- a/ajax/libs/annyang/package.json
+++ b/ajax/libs/annyang/package.json
@@ -24,7 +24,7 @@
     "target": "git://github.com/TalAter/annyang.git",
     "basePath": "dist/",
     "files": [
-      "*"
+      "**/*"
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


ajax/libs/annyang/package.json
```diff
@@ -22,10 +22,9 @@
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/TalAter/annyang.git",
-    "basePath": "",
+    "basePath": "dist/",
     "files": [
-      "annyang.min.js",
-      "annyang.js"
+      "*"
```
PeterDaveHello: @TalAter I'll like to use `**/*` here so that it can also fetch other files under each directories if there are directories in future, what do you think?

TalAter:
@PeterDaveHello Good idea. I just updated it.

Thank you.

PeterDaveHello:
@TalAter Thanks!

PR #9321, opened 2016-10-19
Fix sparseCheckout document example output

TakeshiTseng:
/home/peter/.git/ should be ~/cdnjs


PeterDaveHello:
This is not correct, this is an output message from git which is quoted and will be an absolute path.

PeterDaveHello:
@TakeshiTseng Try:

```sh
cd
git init cdnjs-test
```

PR #9390, opened 2016-10-25
include inlite theme for tinymce

CitySim:
for tinyMCE the "inlite" theme is missing
tinymce docs: https://www.tinymce.com/docs/demo/inlite/


PeterDaveHello:
Thanks @CitySim

Issue #9248, opened 2016-10-11
[intern3] Add missing license info of some libs


x09326:
This issue is for **intern3**, cc #5500 

**Please ref to this issue in your PR comment! and commit message**

These libs don't have `license` field, please add the missing license in their **package.json** to help us maintain these libs. You can find the license information in the **GitHub repo** or **homepage** of these libs.

SPDX: [license-list](https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json)

special case : #9318 

Remaining libs:
- [x] dd_belatedpng
- [x] diff_match_patch

yangwanting:
hashgrid

clairetsai818:
tweet

extend1994:
html5-history-api

NiniHsu:
qwery

jj528s35:
flot

jingwei95:
ng-clip

ko19951231:
noisy

emmy41124:
ng-csv

x112358:
screenfull.js

sajochiu:
ie8

dada0423:
swipe

lcd78706:
galleria

dada0423:
swiper

clairetsai818:
ie8

extend1994:
progress.js

clairetsai818:
ng-clip

jingwei95:
hiw-api

jingwei95:
easeyXDM

extend1994:
wysihtml5

dada0423:
taggd

dada0423:
cufon

clairetsai818:
cubism

kiwi89:
knockout-postbox

lcd78706:
geo-location-javascript

yangwanting:
dragdealer

jj528s35:
ng-csv

yangwanting:
heatmap.js

yangwanting:
vue-smart-table

NiniHsu:
owl-carousel

sajochiu:
taggd

PeterDaveHello:
@x09326 please update the checklist for the added license.

nolsherry:
@PeterDaveHello , I have updated the checklist for the added license.

PeterDaveHello:
@nolsherry Thanks

[...]

PR #9397, opened 2016-10-25
add license of noisy library, cc #9248

ko19951231:
https://github.com/DanielRapp/Noisy/blob/master/MIT-LICENSE.txt


pvnr0082t:
LGTM

PR #9404, opened 2016-10-25
Add a license of ng-clip, cc #9248

clairetsai818:
https://github.com/asafdav/ng-clip/blob/master/LICENSE


PeterDaveHello:
Duplicated with #9407

PR #9420, opened 2016-10-26
Add a license of cubism, cc #9248

clairetsai818:
https://github.com/square/cubism/blob/master/LICENSE

### DIFF
--- a/ajax/libs/cubism/package.json
+++ b/ajax/libs/cubism/package.json
@@ -26,9 +26,9 @@
       "basePath": "",
       "files": [
         "cubism.v1.js",
-        "cubism.v1.min.js",
-        "license": "Square"
+        "cubism.v1.min.js"
       ]
     }
-  ]
+  ],
+  "license": "Square"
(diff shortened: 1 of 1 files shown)


pvnr0082t:
@clairetsai818 You added license at wrong place. That place is still in `npmFileMap` field, but you need to add license as new field.

clairetsai818:
@pvnr0082t okay, thanks.

ajax/libs/cubism/package.json
```diff
@@ -29,5 +29,6 @@
         "cubism.v1.min.js"
       ]
     }
-  ]
+  ],
+  "license": "Square"
```
PeterDaveHello: Square is the company name, not the license.

PeterDaveHello:
Better now but still wrong, @clairetsai818 take a look at the SPDX license list for the valid license identifier:
https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json

Issue #9333, opened 2016-10-20
Merge / migrate tools/fix* programs


PeterDaveHello:
Should merge them into a single `fixFormat.js`

petrgazarov:
Hey @PeterDaveHello! I'm new to the codebase and would like to contribute to this. This task sounds clear to me and looks like a good first-PR.

PeterDaveHello:
@petrgazarov That's nice! Please ping me when you send the PR!

petrgazarov:
@PeterDaveHello I'm trying to clone the repo, but looks like its too big for the amount of memory I have, ~17GB not enough. How large is the repo and is there a way to reduce the clone memory?

PeterDaveHello:
@petrgazarov https://github.com/cdnjs/cdnjs/blob/master/documents/sparseCheckout.md
SparseCheckout and shallow clone will help, thanks!

petrgazarov:
Awesome, I knew there must be a way!

PeterDaveHello:
Done in #9346 but forgot to close issue in commit message, manually close it now. Thanks @petrgazarov

PR #9436, opened 2016-10-28
Merge/migrate tools/fix* programs

petrgazarov:
Addresses https://github.com/cdnjs/cdnjs/issues/9333.
- Consolidates `tools/fixFilenameField.js`, `tools/fixLicense.js` and `tools/fixFormat.js` into `tools/fixFormat.js`.
- removes `fixRepo.js` as it is a duplicate of `fixLicense.js`.

New to the repo, so eager to work with any feedback. Thanks!!


petrgazarov:
@PeterDaveHello Ready for a review

PeterDaveHello:
@petrgazarov Awesome! Thanks!

PeterDaveHello:
@cdnjs/intern2 please take the first review of this change.

pvnr0082t:
LGTM

kennynaoh:
@petrgazarov 
The PR https://github.com/cdnjs/cdnjs/pull/9461 just merge, please fix the conflict.
Thanks!

petrgazarov:
Rebased 🚀

PeterDaveHello:
@cdnjs/intern2 let's review it again!

pvnr0082t:
LGTM

PeterDaveHello:
hmmmm ... actually, I mean to do them all in the single async.each call, which would be more efficient, @petrgazarov would you be interedted in doing that? If not, we can still merge it and do it later, but I'll prefer to do it together, thanks.

petrgazarov:
@Yup, @PeterDaveHello. I just wasn't sure if functions were ever to be used apart from each other. Will make the change.

PeterDaveHello:
@petrgazarov Thank you very much!

petrgazarov:
@PeterDaveHello Done. Please check. I checked that no errors are raised but not sure about the best way to test all uses/edge cases. What do you think about adding a unit test here?

PeterDaveHello:
@petrgazarov It would be very awesome to add aome tests in, we are really lack of testing! I could provide some test cases if you will be interested in! Thanks a lot.

petrgazarov:
Sure, @PeterDaveHello, thanks. I'll try to get to it in the next couple days. I see you have `vows` library already in the code, is that your preference for unit testing?

PeterDaveHello:
@petrgazarov I'm fine if you prefer to use some other test framework which you are familiar with :)

petrgazarov:
Added tests 🚀

PeterDaveHello:
@petrgazarov Awesome! I'll take a look at it, thanks a lot.

PeterDaveHello:
@pvnr0082t let's verify the behavior of this change!

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "git://github.com/cdnjs/cdnjs.git"
   },
   "scripts": {
-    "test": "vows"
+    "test": "vows && jasmine"
```
PeterDaveHello: I wonder if we could separate this part of test to something like `devTest` as we use `npm test` for libraries' package.json validation for a long time, what do you think?

petrgazarov:
Updated, @PeterDaveHello.

tools/fixFormat.js
```diff
@@ -1,104 +1,238 @@
...
+    if (orig[orig.length - 2] !== 'min') {
+      var temp = orig,
+        ext = temp.pop();
+      temp.push('min');
+      temp.push(ext);
+      min = temp.join('.');
+    }
+    if (min !== '' && isThere('./ajax/libs/' + pkg.name + '/' + pkg.version + '/' + min)) {
+      pkg.filename = min;
+      fs.writeFileSync(item, JSON.stringify(pkg, null, 2) + '\n', 'utf8');
```
PeterDaveHello: Looks like we'll write the file at line 36, so we still need to write file here? Thanks.

petrgazarov: Yeah. There are a couple places where this happens. It should probably be done only once, in the end of the function.  I'll check on it.

petrgazarov:
In the recent PR update I optimized the `writeFile`, added 3 more test examples and a hack for the fixFormat function to not be executed for real when running jasmine tests. 🎸  Let me know if this is good.

PeterDaveHello:
@pvnr0082t please also help test this.

pvnr0082t:
@PeterDaveHello I have tested and it can work well. :+1:

PeterDaveHello:
@pvnr0082t Great!

PeterDaveHello:
@petrgazarov can we remove "updated" message - `Library - react-virtualized was updated!` if its `package.json` is not changed? I thought it's the legacy part from `fixLicense.js` ... sorry about that.

petrgazarov:
@PeterDaveHello Good call. I'm checking now with lodash `isEqual` which performs a deepdiff.

PeterDaveHello:
@petrgazarov just noticed that the indent will not be fixed if the json content is not updated, I think we also need to stringify json to compare the indent (or you have some better ideas).

petrgazarov:
@PeterDaveHello Yeah, great point. I would say, if logging which package json files were updated is important, this solution is probably the best bet. Normally if that's not important, I would just write over it regardless, since there isn't much of a performance win imo (considering file is read and diffed in the if condition).

petrgazarov:
Thanks for the detailed code review btw, and if any other such tasks come your way, feel free to tag me!

PeterDaveHello:
@petrgazarov no worry, I think it's time to merge this, we'll see :)

PeterDaveHello:
@petrgazarov would you be interested in helping add the feature to remove the keys with leading underscore? Which are added by the new versions of npm, thanks a lot!

PeterDaveHello:
For example: https://github.com/clairetsai818/cdnjs/blob/64eb46ae547357fcd918d01be5ebaa69b4330d9e/ajax/libs/MinPubSub/package.json#L31-L40

petrgazarov:
@PeterDaveHello Totally!

PeterDaveHello:
@petrgazarov Thanks a lot! (Maybe can even add this in the test!)

PR #9455, opened 2016-11-01
Update package.json for just-animate

notoriousb1t:
- updated website link
- updated folder structure; will be moved to dist from browser folder when I publish 1.1.0
- changed filename to just-animate-all.min.js (the current filename doesn't point to the actual lib)

### DIFF
--- a/ajax/libs/just-animate/package.json
+++ b/ajax/libs/just-animate/package.json
@@ -15,7 +15,7 @@
     "waapi"
   ],
   "license": "MIT",
-  "homepage": "http://just-animate.com",
+  "homepage": "http://just-animate.com/",
   "repository": {
     "type": "git",
     "url": "https://github.com/just-animate/just-animate"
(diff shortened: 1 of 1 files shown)


ajax/libs/just-animate/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "just-animate",
-  "filename": "just-animate-animations.min.js",
+  "filename": "just-animate-all.min.js",
```
PeterDaveHello: Please don't touch filename here unless you added a version with the new name. I can take care of this change, thanks.

notoriousb1t: For some reason I thought it would auto pull from NPM :)  I'll amend this with the 1.0.4 files

notoriousb1t:
I'm going to resubmit this in a new PR.  Sorry for the trouble so far.

PeterDaveHello:
@notoriousb1t No problem, don't worry :)

PR #9652, opened 2016-11-18
Add license info of requirejs-async, cc #5500

clairetsai818:
Reference: https://github.com/millermedeiros/requirejs-plugins/blob/master/LICENSE.txt


clairetsai818:
@PeterDaveHello Please help me review it, thanks.

PR #9653, opened 2016-11-18
Add license info of easyXDM, cc #5500

clairetsai818:
Reference: https://github.com/oyvindkinsey/easyXDM/blob/master/MIT-license.txt


clairetsai818:
@PeterDaveHello Please help me review it, thanks.

PR #9679, opened 2016-11-20
Add license of jquery-sparklines, cc #5500

ChiWeiHsiao:
Reference:
https://github.com/gwatts/jquery.sparkline/blob/master/README.md


PeterDaveHello:
I don't think `New BSD License` is clear enough for us to say that it's BSD-3-Clause, may need to confirm with its author.

PR #9688, opened 2016-11-21
Add license of  dragdealer,cc #9248

yangwanting:
https://github.com/skidding/dragdealer/blob/master/LICENSE
@x09326


x09326:
close due to duplicate with #9459

PR #9723, opened 2016-11-22
add license of webicons, cc #5500

maruilian11:
Reference:
https://github.com/adamfairhead/webicons/issues/2#issuecomment-6919938

### DIFF
--- a/ajax/libs/webicons/package.json
+++ b/ajax/libs/webicons/package.json
@@ -11,5 +11,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/adamfairhead/webicons.git"
-  }
+  },
+  "license": "CC-BY-3.0-US"
 }


PeterDaveHello:
Let's focus on other old request issues, license is not that important and the new interns will take care of it.

maruilian11:
ok, so I just close this?

ajax/libs/webicons/package.json
```diff
@@ -11,5 +11,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/adamfairhead/webicons.git"
-  }
+  },
+  "license": "CC-BY-3.0"
```
PeterDaveHello: What about the US part?

ajax/libs/webicons/package.json
```diff
@@ -11,5 +11,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/adamfairhead/webicons.git"
-  }
+  },
+  "license": "CC-BY-3.0-US"
```
maruilian11: @PeterDaveHello is this format ok?

PeterDaveHello: Do you see this string in the license list?

maruilian11: I only find `CC-BY-3.0` in following license list
https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json#L63
https://spdx.org/licenses/CC-BY-3.0.html

so I need to separate the license to `CC-BY-3.0` and US part?

PeterDaveHello: You need to confirm what's SPDX license id of it ...

PeterDaveHello:
Any updates here? Thanks!

maruilian11:
@PeterDaveHello still get no response from the author.

sufuf3:
@maruilian11 as https://freedomdefined.org/Licenses/CC-BY-3.0 said the https://creativecommons.org/licenses/by/3.0/us/ is for the legal code of the license.
I think many countries use CC-By-3.0 for their legal code of the license, such as Taiwan https://creativecommons.org/licenses/by/3.0/tw/

So I think we can use `CC-BY-3.0` as the license, what do you think?
Thank you.

maruilian11:
@sufuf3 I tried before. Because `CC-BY-3.0-US` doesn't exist in SPDX licenses IDs, we need to confirm  it with the author.
@PeterDaveHello Since the author hasn't replied for a long time, can we add it with `CC-BY-3.0` first?

PeterDaveHello:
I don't believe that we can use `CC-BY-3.0` "because" the author didn't reply, can you please confirm if it make sense to use `CC-BY-3.0` represent the US version first?

maruilian11:
@PeterDaveHello 
https://creativecommons.org/licenses/by/3.0/us/
https://creativecommons.org/licenses/by/3.0/tw/
Seems like they just write in different language to explain the license, so I think we can use `CC-BY-3.0`.

PeterDaveHello:
I don't think they are the same, see the reference:
http://lucien.cc/20150927-3-0-%E5%8F%B0%E7%81%A3%E9%82%84%E6%98%AF-4-0-%E5%9C%8B%E9%9A%9B-%E4%B8%8D%E5%90%8C%E7%89%88%E6%9C%AC%E7%9A%84%E5%89%B5%E7%94%A8cc-%E6%8E%88%E6%AC%8A%E6%A2%9D%E6%AC%BE%E8%A9%B2%E5%A6%82/

PeterDaveHello:
May need more discussions and suggestions, or even asking people good at licensing.

sashberd:
@maruilian11 @PeterDaveHello 
I read this [one ](https://creativecommons.org/licenses/by/3.0/us/deed.ru)of global and this [one ](https://creativecommons.org/licenses/by/3.0/deed.ru)of USA on my native language. The difference is only covering area of license. that is  all.
Suggest to use CC-BY-3.0 (global)
Another way: Why we cannot add another library to our license list?

ajax/libs/webicons/package.json
MattIPv4: This is a valid CC license, just not in the SPDX list.

Issue #8527, opened 2016-07-15
[Request] update viz.js


bfly2000:
viz.js
https://github.com/mdaines/viz.js/
BSD

BigBlueHat:
Looks like it's [The Unlicense](http://unlicense.org)--see https://github.com/mdaines/viz.js/issues/67#issuecomment-255104265 for more.

pvnr0082t:
@PeterDaveHello So we can add this with `"license": "Unlicense"`?

PeterDaveHello:
@pvnr0082t See tools/license-list.json

PR #9724, opened 2016-11-22
[Author] Change viz.js license to MIT

mdaines:
I've decided to license this project under the MIT license instead. See https://github.com/mdaines/viz.js/blob/master/LICENSE and #8527


PeterDaveHello:
Thanks @mdaines

PR #9783, opened 2016-11-29
Add license of foonav, cc #5500

lcd78706:
Ref: https://github.com/fooplugins/foonav/blob/master/dist/foonav.min.js#L6


lcd78706:
@kennynaoh 
Please review this PR. Thanks!

PR #9859, opened 2016-12-04
Add license of LAB.js

hare1039:
ref: https://github.com/getify/LABjs/blob/master/LAB.js#L3


PeterDaveHello:
@hare1039 thanks for your contribution.

Issue #9943, opened 2016-12-12
Remove keys with leading underscore from package.json via fixFormat script


petrgazarov:
Reported by @PeterDaveHello, new versions of npm add extra keys with a leading underscore to package.json.

We should remove those keys in fixFormat script.

PeterDaveHello:
Forgot to close this issue in the PR

PR #9945, opened 2016-12-12
Remove keys with leading underscore from package json

petrgazarov:
Addresses issue: https://github.com/cdnjs/cdnjs/issues/9943

### DIFF
--- a/spec/tools/fixFormatSpec.js
+++ b/spec/tools/fixFormatSpec.js
@@ -922,4 +922,30 @@ describe('fixFormat - tool to fix package.json file of libraries', () => {
       expect(result).toEqual(JSON.stringify(this.packagesMock[0], null, 2) + '\n');
     });
   });
+
+  describe('When there are keys with leading underscores', () => {
+    beforeEach((done) => {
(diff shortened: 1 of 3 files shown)


petrgazarov:
@PeterDaveHello 👓

PeterDaveHello:
@petrgazarov would you like to update the test(npm test) in this same PR? Thanks!

petrgazarov:
@PeterDaveHello Not sure if I understand completely. I thought you didn't want to run this spec on `npm run test`. Isn't it used to run package json validation tests?

PeterDaveHello:
@petrgazarov sorry that I didn't explain clear enough, actually the part you did is the format fixer and its own test, right? The additional part I'm asking is about the package json validation tests - `test/valid-packages-test.js`, make sure the `package.json`s of the libraries with the leading underscore will fail, what do you think?

petrgazarov:
@PeterDaveHello Thanks for pointing that out, I updated the PR. I tested that it prints validation error when I run `vows` with leading underscore keys. Then fixFormat removes them and `vows` passes again.

PeterDaveHello:
@petrgazarov Awesome, thanks a lot.

PR #10428, opened 2017-01-26
Add license of tablefilter, cc #5500

sufuf3:
reference: http://tablefilter.free.fr/


PeterDaveHello:
@sufuf3 let's send another PR to help update this lib, as you can see it's out-dated.

PR #10533, opened 2017-02-04
Fix Vega homepage

domoritz:


PeterDaveHello:
Thanks for your contribution @domoritz

PR #10535, opened 2017-02-04
Update Vega auto-update config for v3

domoritz:
Starting with Vega 3, compiled files are in `/build/`. We are already releasing beta builds. With this change, they will be available on cdnjs.

### DIFF
--- a/ajax/libs/vega/package.json
+++ b/ajax/libs/vega/package.json
@@ -13,8 +13,13 @@
     {
       "basePath": "",
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.+(js|js.map)"
+      ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "vega*.+(js|js.map)"
       ]
     }
   ],


ajax/libs/vega/package.json
```diff
@@ -13,8 +13,13 @@
...
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.+(js|js.map)"
+      ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "vega*.+(js|js.map)"
```
PeterDaveHello: What do you think about just using `"**/vega*"` which can grab files under different directories with the same directory/file structure?

ajax/libs/vega/package.json
```diff
@@ -13,8 +13,13 @@
     {
       "basePath": "",
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.+(js|js.map)"
+      ]
```
PeterDaveHello: for v3, I think maybe we can just remove this filemap with `"basePath": ""` since there is no release file at root, am I right?

PeterDaveHello:
I just manually added v3 beta versions first so that the users can use it on the CDN before we get this merged.

ajax/libs/vega/package.json
domoritz: Yes. However, we may make more v2 patch releases in the future. I just wanted to make sure that cdnjs cover v2 and v3 releases.

ajax/libs/vega/package.json
domoritz: That should be fine. Unfortunately, I had troubled checking out the whole repo (even with partial checkouts) so I had to make my changes using on github online. This makes it hard for me to update this pr. Can you just push these changes directly?

domoritz:
> I just manually added v3 beta versions first so that the users can use it on the CDN before we get this merged.

Fantastic! Thanks.

domoritz:
Cleaned up in https://github.com/cdnjs/cdnjs/pull/10537

ajax/libs/vega/package.json
PeterDaveHello: @domoritz it's fine to make changes online.

PR #10536, opened 2017-02-04
Vega-Lite does not need css files

domoritz:


PeterDaveHello:
@domoritz that's intention to do that for potential future changes compatible, flexibility, so we'll keep it, thanks.

domoritz:
Thanks @PeterDaveHello for looking at my PRs! Vega-Lite will probably never have css files but I'm okay with keeping this as is.

PeterDaveHello:
@domoritz Thank you, too.

PeterDaveHello:
@domoritz We just try to use a simple way that can applied on many libraries so the maintanence work load can be lower, we are only few people and mostly students, but there are 2.8k libraries there, we actually don't have enough support and resource :)

domoritz:
@PeterDaveHello Thanks for your hard work!

PeterDaveHello:
@domoritz you're welcome.

PR #10537, opened 2017-02-04
Update Vega auto-update config for v3 & author info

domoritz:
Supercedes https://github.com/cdnjs/cdnjs/pull/10535

cc @PeterDaveHello

### DIFF
--- a/ajax/libs/vega/package.json
+++ b/ajax/libs/vega/package.json
@@ -13,13 +13,13 @@
     {
       "basePath": "",
       "files": [
-        "vega*.(js|css|map)"
+        "vega*.+(js|css|map)"
       ]
     },
     {
       "basePath": "build",
       "files": [
-        "**/vega*.(js|css|map)"
+        "**/vega*.+(js|css|map)"
       ]
     }
   ],
--- a/ajax/libs/vega/package.json
+++ b/ajax/libs/vega/package.json
@@ -13,8 +13,7 @@
     {
       "basePath": "",
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "**/vega*(js|css|js.map)"
       ]
     }
   ],
@@ -24,7 +23,7 @@
   },
   "license": "BSD-3-Clause",
   "author": {
-    "name": "Arvind Satyanarayan",
-    "url": "http://arvindsatya.com"
+    "name": "UW Interactive Data Lab",
+    "url": "http://idl.cs.washington.edu"
   }
 }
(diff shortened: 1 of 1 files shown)


ajax/libs/vega/package.json
```diff
@@ -13,8 +13,7 @@
     {
       "basePath": "",
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "**/vega*(js|css|js.map)"
```
PeterDaveHello: I think its should be `vega*.(js|css|map)`, `**/` is not allowed in root path.

PeterDaveHello: So you'll still need to add another filemap here.

domoritz: Done.

ajax/libs/vega/package.json
```diff
@@ -13,8 +13,13 @@
...
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.(js|css|map)"
+      ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "vega*.(js|css|map)"
```
PeterDaveHello: As we discussed earlier, maybe you'll like to use `"**/vega*.(js|css|map)"` here?

domoritz: Oh okay. I can change it but with the basePath set to build I suspected this would be sufficient for now and into the future.

domoritz: Done

ajax/libs/vega/package.json
```diff
@@ -13,8 +13,13 @@
...
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.(js|css|map)"
+      ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "**/vega*.(js|css|map)"
```
PeterDaveHello: Sorry that I didn't note that the syntax is wrong, should be `"**/vega*.+(js|css|map)"` (needs **+**)

ajax/libs/vega/package.json
```diff
@@ -13,8 +13,13 @@
     {
       "basePath": "",
       "files": [
-        "vega.js",
-        "vega.min.js"
+        "vega*.(js|css|map)"
```
PeterDaveHello: Sorry that I didn't note that the syntax is wrong, should be `vega*.+(js|css|map)"` (needs **+**)

domoritz: Fixed

ajax/libs/vega/package.json
domoritz: Fixed

PR #10562, opened 2017-02-08
[author] Connect openlayers to npm

ahocevar:
Related issue: #69

OpenLayers is on npm, and version 4.0.0 is about to be released. There is also `ol3` here, but that implies version 3.x, which will be a bit weird to go on when major releases are >3.


ahocevar:
Created a new, clean pull request for this at #10563.

PR #10760, opened 2017-03-01
Update auto-update config of Vega-Lite

domoritz:
https://github.com/vega/vega-lite/pull/1957

### DIFF
--- a/ajax/libs/vega-lite/package.json
+++ b/ajax/libs/vega-lite/package.json
@@ -16,6 +16,12 @@
       "files": [
         "vega-lite*.+(js|css|map)"
       ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "**/vega-lite*.+(js|css|map)"
+      ]
     }
   ],
   "keywords": [


PeterDaveHello:
@domoritz Please use a meaningful PR title with the library name, thanks.

domoritz:
> Should we add this? It seems like we should modify the old one?

I'd like to stay backwards compatible. We may re-release a 1.x version. 

> Also, why .css and .map?

For the future as suggested by @PeterDaveHello. Vega uses the same format.

ajax/libs/vega-lite/package.json
```diff
@@ -16,6 +16,12 @@
       "files": [
         "vega-lite*.+(js|css|map)"
       ]
+    },
+    {
+      "basePath": "build",
+      "files": [
+        "**/vega-lite*.+(js|css|map)"
```
pvnr0082t: @domoritz I'd like to make sure If there are only distributed purpose files under `build` directory. If so, we can use `"**/*"` instead. If not, let's just keep this.

domoritz: We build every single `.ts` file and put it in `build`. I don't think they should be on a cdn.

pvnr0082t: @domoritz Okay, let's keep it. :+1:

domoritz:
@PeterDaveHello Let me know if you want any additional changes.

Issue #10713, opened 2017-02-24
Update dragdealer


PeterDaveHello:
The latest version seems to be v0.9.9: https://github.com/skidding/dragdealer/tree/0.9.9, but we only have v0.9.8 on https://cdnjs.com/libraries/dragdealer

PeterDaveHello:
Ping @maruilian11

PR #10875, opened 2017-03-18
Update dragdealer package.json info, add auto-update and old versions files

maruilian11:
@cdnjs/intern2 Please help me review this PR, thanks.
The author only provide dragdealer.min.js in 0.9.6 ~ 0.9.8, so I manually add them.
Pull request for issue: #10713

 * GitHub / Bitbucket popularity (required):
   Watch 39
   Star 806
   Fork 134
   60 downloads in the last day
   230 downloads in the last week
   1,475 downloads in the last month


maruilian11:
@pvnr0082t the author do provide the minified files
https://github.com/skidding/dragdealer/blob/0.9.6/lib/dragdealer.min.js
https://github.com/skidding/dragdealer/blob/0.9.7/lib/dragdealer.min.js


>In e192284, I think update keywords and add homepage and author info should be separate into two commits.

I think they are all belong to updating the package.json, I think it's fine, or we can ask the manager's opinion, thanks.

maruilian11:
@sufuf3 I mention the minify tool, which means if I minify the file, the tool I used would be...

Put 
```
Source: git repo
Js minify tool: uglify-js
Css minify tool: clean-css
```
between 0.9.5 and 0.9.6~0.9.7 seem more weird? 

P.S. Which line you think isn't a sentence?

sufuf3:
I think the following part is for v0.9.5~v0.9.7, but actually, it's only for v0.9.5.
```
Source: git repo
Js minify tool: uglify-js
Css minify tool: clean-css
```
How about
```
Add dragdealer old versions manually, cc #10713

- Add 0.9.5 from GitHub due to the different file structure with current 
  auto-update. Minify the files by uglify-js & clean-css.
- Add 0.9.6 and 0.9.7 because our current auto-update can't fetch
  dragdealer.min.js in both of them.
```
?

What do you think?

maruilian11:
@sufuf3 Is that ok?

maruilian11:
@extend1994 
1. I think we only use `with` when adding versions. Like `Add xxx old versions...`, we won't use `Add xxx with old versions... `, right?
2. Thanks for your opinion, but prefer separate them.
3. Updated, thanks.

PR #10980, opened 2017-04-02
jquery.tablesorter: Fix SPDX license. See #5544

Mottie:


ajax/libs/jquery.tablesorter/package.json
```diff
@@ -17,10 +17,7 @@
     "alphanumeric",
     "natural"
   ],
-  "licenses": [
-    "MIT",
-    "deprecated_GPL-2.0+"
-  ],
+  "licenses": "(MIT OR GPL-2.0+)",
```
PeterDaveHello: sorry but we don't support this format but use the array

Mottie: Oh, I thought you wanted a SPDX string... I should have used "license" (no "s").

PeterDaveHello:
Though http://git.spdx.org/ is removed ... the last time we synced from upstream (SPDX v2.5) it's:
https://github.com/cdnjs/cdnjs/blob/12d79a4bf555f80bd5b8c1b694ccf127e69aea0c/tools/license-list.json#L113
`  "deprecated_GPL-2.0+",`

ajax/libs/jquery.tablesorter/package.json
Mottie: If it doesn't matter, then we can just close this PR.

PeterDaveHello: but that's not how we use it currently, hmmm ...

PeterDaveHello: I think it doesn't really matter, if there is any inconvenience caused, please feel free to discuss with us, thanks!

PR #11035, opened 2017-04-09
Remove unrecognized keys in fixFormat, closes #5452

### LINKED ISSUE #5452: Deny all the un-recognized fields in package.json?
Maybe we can deny all the fields that we cannot recognize in package.json so that we won't have this problem again?
https://github.com/cdnjs/cdnjs/commit/b435c303c5afe09de042eff3982e5670163c68d5

@cdnjs/intern @cdnjs/team-cdnjs

petrgazarov:
Pull request for issue: #5452 

# Git commit checklist
 * [x] The first line of commit message is less then 50 chars, be clean and clear, easy to understand.
 * [x] The parent of the commit(s) in the PR is not old than 3 days.
 * [x] Pull request is sending from a non-master branch with meaningful name.

___________________________________________________________________________________

Removes all unsupported keys from package.json when running `fixFormat`.
I looked at our existing validations, as well as cndjs-importer to figure out which fields we need to leave. If I missed some, please point it out. Thanks 😄

### DIFF
--- a/ajax/libs/EventEmitter/package.json
+++ b/ajax/libs/EventEmitter/package.json
@@ -19,7 +19,6 @@
     "url": "http://oli.me.uk/"
   },
   "license": "Unlicense",
-  "readmeFilename": "README.md",
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/Olical/EventEmitter.git",
--- a/ajax/libs/angucomplete-alt/package.json
+++ b/ajax/libs/angucomplete-alt/package.json
@@ -4,7 +4,6 @@
   "version": "3.0.0",
   "description": "Awesome Autocompleteness for AngularJS",
   "author": "ghiden",
-  "email": "hidenari@gmail.com",
   "repository": {
     "type": "git",
     "url": "https://github.com/ghiden/angucomplete-alt.git"
--- a/ajax/libs/angular-cache/package.json
+++ b/ajax/libs/angular-cache/package.json
@@ -23,7 +23,6 @@
       ]
     }
   ],
-  "postinstall": "node node_modules/grunt-cli/bin/grunt build",
   "license": "MIT",
   "keywords": [
     "CacheFactory",
--- a/ajax/libs/angular-gantt/package.json
+++ b/ajax/libs/angular-gantt/package.json
@@ -15,7 +15,6 @@
     "chart",
     "angularjs"
   ],
-  "directories.doc": "docs",
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/angular-gantt/angular-gantt.git",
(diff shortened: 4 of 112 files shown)


petrgazarov:
cc/ @PeterDaveHello @maruilian11 @IonicaBizau 

Hey guys, sorry didn't realize github will send notification before PR is opened. 👀 please

PeterDaveHello:
Looks awesome, I think it'll be even better if the coding style things can be separated into a different commit next time.

petrgazarov:
@PeterDaveHello Yeah! Totally forgot about the test.

PeterDaveHello:
@petrgazarov and if it failed, please also come up with a commit to fix the broken package.json :)

petrgazarov:
@PeterDaveHello Ah, I thought it's better to have one commit per change in the PR? Don't want to polute your commit history by adding commits that fix the previous, unmerged commit.

PeterDaveHello:
@petrgazarov I think it depends, they actually are really related, just like you can put the fix tool and the test in this same PR,

petrgazarov:
@PeterDaveHello I'll take a  look at the CI error tomorrow.  I think that with all the  libraries already there, there will have to be some corrections to their package.json files.  My understanding is that the test runs on all of them in CI? I'm not sure if I have enough CPU or memory to check out the entire repository.  Any insight  of a good way to update them?

PeterDaveHello:
@petrgazarov yeah all of them, if there is problem for you, I can update it.

petrgazarov:
Hey @PeterDaveHello sorry it take me a while to get to this, but should be viewable now. Because of memory constraints, I had to upload libs in multiple commits, then fixup all at the end. Something I noticed while running `fixFormat` -- a lot of libs have a problem with license (either don't have it, or have unrecognized one). Perhaps we should make it a required field in the test, and at the same time, go through the missing ones, and fill them one by one. Thoughts?

PeterDaveHello:
@petrgazarov thanks a lot, and sorry for the memory issue.

About the license issue, we are still trying to make them correct, as you can see there is progress here : #5500 #6324, but I have to say that it is really a hard work, some libraries are not under maintenance any more, some of them are even deleted from upstream, add it to the test is exactly what I want to do for years but still trying, now we can only make sure the new added libraries contain the correct license manually 😢

PeterDaveHello:
@petrgazarov sorry that I may be a little bit late, I wonder if we can temporarily keep some of the keys:
 - title
 - namespace

This PR and the code inside it it's very great and helpful 👍

petrgazarov:
Hey @PeterDaveHello, I added the two fields you mentioned to the list of recognized fields. Take a look when you have a chance. Thanks!

PeterDaveHello:
@petrgazarov thanks!

PR #11088, opened 2017-04-17
Simplify file map of Genericons

jdh8:
Update of Genericons seems to be stopped.  Perhaps this would help.

### DIFF
--- a/ajax/libs/genericons/package.json
+++ b/ajax/libs/genericons/package.json
@@ -20,9 +20,9 @@
     "target": "git://github.com/Automattic/Genericons.git",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "genericons",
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
The version on CDNJS is the latest release: https://github.com/Automattic/Genericons/releases

jdh8:
From 3.2 on they don't tag but branch new releases.  There are stable `release/3.2` and `release/3.3` branches which are not modified since 2014.  How does our autoupdater work?

Because direct merge from `master` is prohibited, I'll start over a new pull request on a differently named branch.

PeterDaveHello:
@jdh8 we do not use branch as version reference, the versions need to be tagged, once it's tagged, our auto-updater will grab it, thanks!

PR #11208, opened 2017-05-09
[redux-saga] update package.json

Andarist:


ajax/libs/redux-saga/package.json
```diff
@@ -16,8 +16,11 @@
     "side effects"
   ],
   "author": "Yassine ELOUAFI <yelouafi@gmail.com>",
+  "contributors": [
+    "Mateusz Burzyński <mateuszburzynski@gmail.com> (https://github.com/Andarist)"
+  ],
```
PeterDaveHello: Would you please remove `contributors` since we don't use it. Thanks.

Andarist: @PeterDaveHello done, just squash those 2

PeterDaveHello:
Thanks @Andarist

PR #11211, opened 2017-05-10
Add license of PKI.js

kweisamx:
ref: https://github.com/PeculiarVentures/PKI.js/blob/master/LICENSE, https://github.com/PeculiarVentures/PKI.js#license

### DIFF
--- a/ajax/libs/PKI.js/package.json
+++ b/ajax/libs/PKI.js/package.json
@@ -30,5 +30,6 @@
         ]
       }
     ]
-  }
+  },
+  "license": "BSD-3-Clause"
 }


ajax/libs/PKI.js/package.json
```diff
@@ -30,5 +30,6 @@
         ]
       }
     ]
-  }
+  },
+  "license": "MIT"
```
pvnr0082t: It seems like it's `BSD-3-Clause`. 
ref: https://github.com/PeculiarVentures/PKI.js/blob/master/LICENSE, https://github.com/PeculiarVentures/PKI.js#license

ajax/libs/PKI.js/package.json
```diff
@@ -30,5 +30,6 @@
         ]
       }
     ]
-  }
+  },
+  "license": "BSD-3-clause"
```
sufuf3: `BSD-3-Clause`
Please use upper case. c -> C
Let it as same with https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json#L49

kweisamx: thank !

PeterDaveHello:
@kweisamx I think you can try to send a pull request to update its repository url.

PR #11228, opened 2017-05-11
Update repository url, homepage, keyword info of PKI.js

kweisamx:
Pull request for issue: #11268
ref : https://github.com/PeculiarVentures/PKI.js/blob/master/package.json#L15
https://github.com/PeculiarVentures/PKI.js/blob/master/package.json#L54-L67
https://pkijs.org/

### DIFF
--- a/ajax/libs/PKI.js/package.json
+++ b/ajax/libs/PKI.js/package.json
@@ -5,18 +5,23 @@
   "version": "1.2.1",
   "repository": {
     "type": "git",
-    "url": "git://github.com/GlobalSign/PKI.js.git"
+    "url": "git://github.com/PeculiarVentures/PKI.js.git"
   },
-  "homepage": "http://pkijs.org",
+  "homepage": "https://pkijs.org",
   "keywords": [
-    "encryption",
-    "certificate",
+    "ES6",
+    "ES2015",
     "WebCrypto",
-    "PKCS",
-    "ANS.1",
-    "CMS",
+    "Web Cryptography API",
+    "X.509",
+    "certificate",
+    "crl",
+    "cms",
+    "encrypted",
+    "enveloped",
     "OCSP",
-    "x509"
+    "timestamp",
+    "PKCS"
   ],
   "filename": "x509_simpl.min.js",
   "autoupdate": {


ajax/libs/PKI.js/package.json
```diff
@@ -5,7 +5,7 @@
   "version": "1.2.1",
   "repository": {
     "type": "git",
-    "url": "git://github.com/GlobalSign/PKI.js.git"
+    "url": "git://github.com/PeculiarVentures/PKI.js.git"
   },
   "homepage": "http://pkijs.org",
```
sufuf3: How about also update the homepage?
`https://pkijs.org/`, using https is better because it's more security for browsing. To prevent the possibility the users been attacked during the redirection.

ajax/libs/PKI.js/package.json
```diff
@@ -5,7 +5,7 @@
...
   "keywords": [
```
sufuf3: How about also update the keywords, to let it as same with upstream.
ref: https://github.com/PeculiarVentures/PKI.js/blob/master/package.json#L54-L67

kweisamx: ok thanks

kweisamx:
@extend1994 Yes,thank you, but I think if I just say that according to package.json is not preciser , because there are only three place fixed and it's more clearly.

extend1994:
@kweisamx Okay, I think we can wait for others' opinion about this :smiley:

kweisamx:
Sorry,i don't understand what do you mean? Can tell me in detail. Thank you!

PeterDaveHello:
You only mention the info in comment but not in commit message, should better do that in both commit message and comment.

Also, the first line of the commit message should be shorter(better to be within 50 chars). 
Maybe you'd like to take a look at https://github.com/blog/926-shiny-new-commit-styles as reference

PeterDaveHello:
Commit message should be more well formatted, see the revised version here: https://github.com/cdnjs/cdnjs/commit/c4056cefaec9b7aa75492f60ae39cfac696b9ead

PR #11247, opened 2017-05-14
Add license of jquery.sparkline, cc #5500

kweisamx:
ref: https://github.com/gwatts/jquery.sparkline/blob/master/src/header.js#L16
http://omnipotent.net/jquery.sparkline/#s-about , the link of "New BSD License"


kweisamx:
sorry , i sent wrong PR , i try to fix later

PR #11571, opened 2017-07-18
Update typed.js autoupdate config

euler0:
The base path has changed from `dist` to `lib` since v2.0


PeterDaveHello:
Thanks @euler0 @dakshshah96

Issue #11931, opened 2017-10-03
Add/Update a few fields in package.json


extend1994:
This issue is mainly for Hacktoberfest hackathon on Oct 2 2018 in NCTU, Taiwan.

The text you see is the library name in cdnjs, the link is directing you to find the related info.
## Add/Update license field
- Reference license lists:
  - [spdx license list](https://spdx.org/licenses/)
  - [cdnjs license list](https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json)
- [x] [ekko-lightbox](https://github.com/ashleydw/lightbox/blob/master/LICENSE#L1)
- [x] [embedly-jquery](https://github.com/embedly/embedly-jquery/blob/master/LICENSE)
- [x] [equalize.js](https://github.com/tsvensen/equalize.js#legal)
- [x] [intercom.js](https://github.com/diy/intercom.js#license)
- [x] [jo](https://github.com/davebalmer/jo/blob/ec767af487b8653778c36168c7b3a5f478753cc5/LICENSE.mdown) BSD-2-Clause
- [x] [jqPlot](https://github.com/jqPlot/jqPlot/#legal-notices)
- [x] [jquery-form-validator](http://www.formvalidator.net/#license)
- [x] [jquery-mockjax](https://github.com/jakerella/jquery-mockjax#license)
- [ ] [jquery.blockUI](https://github.com/malsup/blockui#copyright-and-license)
- [x] [jquery.collapsible](https://github.com/juven14/Collapsible/blob/master/LICENSE)
- [x] [jquery.form](https://github.com/jquery-form/form#license)
- [x] [jquery.mb.YTPlayer](https://github.com/pupunzi/jquery.mb.YTPlayer/tree/master/licences)
- [x] [machina.js](https://github.com/ifandelse/machina.js/blob/master/LICENSE)
- [ ] [placeholder-shiv](https://github.com/walterdavis/placeholder-shiv/blob/master/LICENSE.txt)
- [x] [jquery.tablesorter](https://github.com/mottie/tablesorter/#licensing)
- [ ] [webkit.js](https://webkit.org/licensing-webkit/) https://github.com/trevorlinton/webkit.js#License
- [ ] [fatcow-icons](http://www.fatcow.com/free-icons) See Terms of Use Creative Commons Attribution 3.0 License -> CC-BY-3.0
- [ ] [log4javascript](http://log4javascript.org/)
[...]

PR #11939, opened 2017-10-03
Update license of yepnope, cc #11931

yaxauw:
ref:
https://github.com/SlexAxton/yepnope.js/blob/master/LICENSE.md


extend1994:
There is another one send a same pull request before you. So... you need to pick another one in the issue. Don't forget you need a space between the word after comma and the comma itself when you commit. And leave a comment about which library you picked in the issue. Thanks!

PR #11942, opened 2017-10-03
Update license information of jit, cc #11931

ralphchung:
ref: https://github.com/philogb/jit/blob/master/LICENSE


extend1994:
Oh... I have to wait for ci test pass.

PR #11944, opened 2017-10-03
Update ionicons license info in package.json

Billy4195:
Update it because it didn't contain `url` field, cc #11931.
Ref: https://github.com/ionic-team/ionicons/blob/master/LICENSE

### DIFF
--- a/ajax/libs/ionicons/package.json
+++ b/ajax/libs/ionicons/package.json
@@ -22,9 +22,7 @@
       }
     ]
   },
-  "license": {
-    "name": "MIT"
-  },
+  "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


ajax/libs/ionicons/package.json
```diff
@@ -23,7 +23,8 @@
     ]
   },
   "license": {
-    "name": "MIT"
+    "name": "MIT",
+    "url": "https://github.com/ionic-team/ionicons/blob/master/LICENSE"
```
sufuf3: Let's modify from A to B to fit the CDNJS pattern.
- A
```
  "license": {
    "name": "MIT",
    "url": "https://github.com/ionic-team/ionicons/blob/master/LICENSE"
  },
```

- B
```
  "license": "MIT",
```

Thank you.

PR #11954, opened 2017-10-04
Add roundabout license info in package.json

finch185277:
cc #11931
Update it because the license roundabout used changed
Ref: https://github.com/fredleblanc/roundabout/blob/master/jquery.roundabout.js#L11-L40

### DIFF
--- a/ajax/libs/roundabout/package.json
+++ b/ajax/libs/roundabout/package.json
@@ -19,5 +19,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/fredleblanc/roundabout.git"
-  }
+  },
+  "license": "BSD-3-Clause"
 }


finch185277:
I did:
 "license":"BSD-3-Clause",  >>>  "license": "BSD-3-Clause",
but ... I still don't know why FAILED
??

finch185277:
(I add a space between : and ".
Like from A to B.

A
"license":"BSD-3-Clause",
B
"license": "BSD-3-Clause",
)

ajax/libs/roundabout/package.json
```diff
@@ -19,5 +19,6 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/fredleblanc/roundabout.git"
-  }
+  },
+  "license": "BSD-3-Clause",
```
extend1994: comma "," have to be removed. It will be
```
"license": "BSD-3-Clause"
```
We don't need comma when there is no other fields.
For example, after "repository", there is still "license", so we need comma to separate them. But there is no field after "license", therefore comma is not needed.

You don't have to open an new pull request when you do the modification, just do it on the same branch of this pull request. In you case, it's `update-roundabout` branch. Update it and commit again. For your reference, commit message can be 
```
Remove spare comma to add license correctly, cc #11931
```
s

PeterDaveHello:
@extend1994 note the PR title.

extend1994:
ping @cdnjs/intern3

sufuf3:
There is no license info in package.json in original package.json https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/roundabout/package.json
So the commit message should be used `Add license info`
I helped modify the commit message and squash the commits.

Thanks

PeterDaveHello:
I think @extend1994 can take care of this one with her merge permission.

PR #11989, opened 2017-10-13
Update antimoderate license and keywords

programmiri:
Pull request for issue: 11931

### DIFF
--- a/ajax/libs/antimoderate/package.json
+++ b/ajax/libs/antimoderate/package.json
@@ -16,15 +16,15 @@
       ]
     }
   ],
-  "license": {
-    "type": "See README.md",
-    "url": "https://github.com/whackashoe/antimoderate#license"
-  },
+  "license": "MIT",
   "author": {
     "name": "Jett LaRue",
     "url": "https://jettlarue.com"
   },
   "keywords": [
-    "antimoderate"
+    "antimoderate",
+    "image",
+    "progressive",
+    "loading"
   ]
 }


PeterDaveHello:
Squashing into only one commit now.

ajax/libs/antimoderate/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "antimoderate",
   "version": "0.0.2",
-  "description": "The progressive image loading library for great good!",
-  "filename": "antimoderate.min.js",
+  "description": "The progressive image loading library",
+  "filename": "antimoderate.js",
```
sufuf3: Let's use original `antimoderate.min.js`.
We also host `antimoderate.min.js` file, so we have to use `antimoderate.min.js` as filename.
Thanks.

ajax/libs/antimoderate/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "antimoderate",
   "version": "0.0.2",
-  "description": "The progressive image loading library for great good!",
-  "filename": "antimoderate.min.js",
+  "description": "The progressive image loading library",
```
sufuf3: It's fine to use origin description.
This is the same with https://github.com/whackashoe/antimoderate#antimoderate
Thanks.

ajax/libs/antimoderate/package.json
```diff
@@ -16,15 +16,13 @@
...
-  },
+  "license": "MIT",
   "author": {
     "name": "Jett LaRue",
     "url": "https://jettlarue.com"
   },
   "keywords": [
-    "antimoderate"
+    "antimoderate",
+    "image"
```
PeterDaveHello: Maybe add some more like `progressive`, `loading`?

ajax/libs/antimoderate/package.json
```diff
@@ -16,15 +16,13 @@
       ]
     }
   ],
-  "license": {
-    "type": "See README.md",
-    "url": "https://github.com/whackashoe/antimoderate#license"
-  },
+  "license": "MIT",
```
PeterDaveHello: From the readme, looks like `StackBlur Algorithm` is under MIT but its self may be CC-0 or public domain?

PeterDaveHello: Though it is MIT in bower.json, but it'll be great that if you can open an issue ticket to confirm with its author about the license, would you?

ajax/libs/antimoderate/package.json
PeterDaveHello: PS: You can also contribute this back to its own bower.json in its repository, or even help it create a package.json if its author accepts.

ajax/libs/antimoderate/package.json
programmiri: Sure, I made an issue in the repo there to confirm the license -> https://github.com/whackashoe/antimoderate/issues/15

maruilian11:
Hi @programmiri, Due to the author's reply, I think we can use `MIT`. :+1: 
Would you please help add more keywords, like `progressive` and `loading`? Thanks.
 https://github.com/cdnjs/cdnjs/pull/11989/commits/9c61f6f1467234ddac989fcc27a01fbc31ccf85b#r145030225

PR #12037, opened 2017-10-23
Request to update library files for new version. Updated package.info file

lokesh-coder:
New version of this library has different filename and base path. Updated version, filename, desc, keywords and npmFileMap.


lokesh-coder:
I will fix them and reopen.

PR #12061, opened 2017-10-26
Add license format of legojs

kasony-zkcodeveryway:
Ref:https://github.com/ahager/legojs/blob/master/LICENSE.txt
cc #11931

### DIFF
--- a/ajax/libs/legojs/package.json
+++ b/ajax/libs/legojs/package.json
@@ -24,4 +24,5 @@
   "keywords": [
     "legojs"
   ]
+  "license": "MIT"
 }


ajax/libs/legojs/package.json
```diff
@@ -24,4 +24,5 @@
   "keywords": [
     "legojs"
   ]
+  "license": "MIT"
```
sufuf3: Duplicated with Line 7
@kasony-zkcodeveryway Would you chose another one for updating a few fields?
Really sorry for not notice about this. cc @extend1994 

If you have any question, please let me know.
I will do my best to help you to make a good PR.

Thank you.

PS1. Because this lib don't have to be update the license. I'll close it.
PS2. @cdnjs/intern3 We have to help update the auto-update config, if you have time, please help make this library better. I can do it recently, after I review the PRs.

PR #12174, opened 2017-11-13
Fix Autoupdater for emoji-picker

hebbet:


ajax/libs/emoji-picker/package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/one-signal/emoji-picker.git",
+    "target": "git://github.com/onesignal/emoji-picker.git",
```
PeterDaveHello: I think this should be `git://github.com/OneSignal/emoji-picker` ?

sashberd: Agreed with @PeterDaveHello

maruilian11:
@hebbet Would you please modify your commit as Peter said?

hebbet:
changed.

Issue #12336, opened 2017-12-15
[Request] Update store.js to v2


andronocean:
[store.js](https://github.com/marcuswestin/store.js) is currently at v2.0.12, while the CDN only has it [up to v1.3.20](https://cdnjs.com/libraries/store.js) from back in 2015. Pretty popular project, and I can't find any other trustworthy CDNs that carry it.

sufuf3:
@andronocean Many thanks.
PS. Now, the bathPath is https://github.com/marcuswestin/store.js/tree/master/dist
https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/store.js/package.json#L14-L27

ralphchung:
I'll take this!

PR #12351, opened 2017-12-19
Update auto-update config of store.js

ralphchung:
Related issue: #12336
The current version on CDNJS is outdated because the file structure on the upstream repo changed. I made this change so that auto-update can successfully grab the latest version.

### DIFF
--- a/ajax/libs/store.js/package.json
+++ b/ajax/libs/store.js/package.json
@@ -16,11 +16,9 @@
     "target": "git://github.com/marcuswestin/store.js.git",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "store.js",
-          "store.min.js",
-          "store+json2.min.js"
+          "**/*"
         ]
       }
     ]


ajax/libs/store.js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "store.js",
-  "filename": "store.min.js",
+  "filename": "store.legacy.min.js",
```
PeterDaveHello: This can't be updated here.

ajax/libs/store.js/package.json
```diff
@@ -16,11 +16,9 @@
     "target": "git://github.com/marcuswestin/store.js.git",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "store.js",
-          "store.min.js",
-          "store+json2.min.js"
+          "**/*.js"
```
PeterDaveHello: Should be `**/*` here.

ajax/libs/store.js/package.json
```diff
@@ -16,11 +16,9 @@
     "target": "git://github.com/marcuswestin/store.js.git",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "store.js",
-          "store.min.js",
-          "store+json2.min.js"
+          "**/*"
```
sufuf3: Ummm ...
I think the files can be used in web front-end is `*.min.js` is lib, because the lib use browserify & uglify-js to generate [`*.min.js`](https://github.com/marcuswestin/store.js/blob/v2.0.12/scripts/compile-builds.js#L39) files.
Also , the README said usage way in https://github.com/marcuswestin/store.js/tree/v2.0.12#installation, it's just mention the minified flies.
And we just need to add https://github.com/marcuswestin/store.js#list-of-default-builds these files, I think add `store.tests.min.js` in CDNJS is a little bit weird, also there is no tutorial mention that we need it.
What do you think?

Many thanks.

PS. Need to use excluded mode in `dist` as the basePath.

sufuf3:
After update the auto-update config.
Would you add the newest version in `second commit`?
This is useful to maintain this lib, also can update the filename and version fields in package.json.

Many thanks.

ajax/libs/store.js/package.json
ralphchung: @sufuf3 
Actually, I don't know...
I still can not get a clear view of the principles that one should follow in the cdnjs.
This rule was actually suggested by the PeterDaveHello as you can see above.
I agree with you thoughts but I just don't know what should I follow.

Thanks!

sufuf3: OK.
Let's wait for @PeterDaveHello 's decision.

sufuf3: @PeterDaveHello would you please help this pattern problem?
You said the pattern should be used `**/*` https://github.com/cdnjs/cdnjs/pull/12351#discussion_r157731405

But I think adding `store.tests.min.js` in CDNJS is a little bit weird, becuase there is no tutorial mention that we need it.
What do you think?

Many thanks.

sufuf3:
@PeterDaveHello would you please help this pattern problem?
You said the pattern should be used `**/*` https://github.com/cdnjs/cdnjs/pull/12351#discussion_r157731405

But I think adding `store.tests.min.js` in CDNJS is a little bit weird, becuase there is no tutorial mention that we need it.
What do you think?

Many thanks.

PeterDaveHello:
@sufuf3 it's minified which is usually for productoin, I'm good if you can confirm about that concern ;)

sufuf3:
@PeterDaveHello  as the comments mention https://github.com/marcuswestin/store.js/blob/master/dist/store.tests.js#L1-L2
```
// store-test-suite will run all the store.js tests
// and report the results.
```
Maybe this file just for the tests.

Also, I noticed the `store.tests.min.js` file is not in https://github.com/marcuswestin/store.js#list-of-default-builds 

Are these information good?
Or, maybe I need to open an issue to discuss the author with this question on the lib's repo.

Thank you.

PeterDaveHello:
@sufuf3 of course a discussion to confirm that will be even better 👍

sashberd:
@MattIPv4 Here we need to exclude *.test.* files from dist folder. As I know we use only **/* glob for dist folder. So I need you suggestion for this one

PR #12403, opened 2018-01-03
Update TableDnD auto-update to only fetch right files

extend1994:
Current version on cdnjs is v0.9.1.
v0.9.2 released 7 days ago, which has filename including version number. 
See https://github.com/isocra/TableDnD/tree/master/dist.
Therefore it's necessary to update auto-update config.


PeterDaveHello:
Note that the expression / glob pattern could be improved to be more precisely.

Issue #12441, opened 2018-01-20
[Request] Please update startupbootstrap-freelancer to its latest version


caiobegotti:
Please update startupbootstrap-freelancer to its latest version: 4.0.0

MattIPv4:
5.x is now available: https://cdnjs.com/libraries/startbootstrap-freelancer

PR #12442, opened 2018-01-20
Update startbootstrap-freelance package.json

caio1982:
The startupbootstrap-freelancer theme has been recently updated to cope with the new 4.0.0 version of Bootstrap itself.

Pull request for issue: #12441

### DIFF
--- a/ajax/libs/startbootstrap-freelancer/package.json
+++ b/ajax/libs/startbootstrap-freelancer/package.json
@@ -1,7 +1,7 @@
 {
   "name": "startbootstrap-freelancer",
   "filename": "js/freelancer.min.js",
-  "version": "3.3.7",
+  "version": "4.0.0",
   "homepage": "https://startbootstrap.com/template-overviews/freelancer/",
   "description": "A one page HTML theme for freelancers created by Start Bootstrap",
   "keywords": [


Billy4195:
The version 4 of startbootstrap-freelancer is release on npm package, not GitHub repo, the auto-update configure of startbootstrap-freelancer in `package.json` using GitHub repo to download source code, you should change it to npm auto-update config.
Refer: [auto-update](https://github.com/cdnjs/cdnjs/blob/master/documents/autoupdate.md)

ajax/libs/startbootstrap-freelancer/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "startbootstrap-freelancer",
   "filename": "js/freelancer.min.js",
-  "version": "3.3.7",
+  "version": "4.0.0",
```
PeterDaveHello: Thanks for your contribution, however, we can not directly bump the version number, if the auto-update didn't fetch the new versions properly, we should try to dig about the root cause, usually to update the auto-update config. If we really need to add the new version manually, we should also commit the assets.

sashberd: @PeterDaveHello There are 3 versions 4.0.0 on npm (alpha, beta and release). Should we add all of the manually or only 4.0.0 (release)?
Also may be we can ask author to add git tag of v4.0.0 to git repo and do nothing :)

sashberd:
@caio1982 The v5.0.0 was tagged and hosted by cdnjs.

Issue #12459, opened 2018-01-24
Wrong version for LiveScript 1.5.0 (it is 1.4.0 on CDN)


pepkin88:
LiveScript seems to have configured an auto-updater, as of https://github.com/cdnjs/cdnjs/pull/3897, but the newest browser build was generated right after the 1.5.0 tagged commit (and was fixed later https://github.com/gkz/LiveScript/commit/87daae63c4e988eb5c5d0ab599654f58af6c8d0a).

Is there a way to trigger a reupload of the newest version? I don't have rights to the LiveScript repository.

These are the correct files: https://github.com/gkz/LiveScript/tree/master/browser

extend1994:
Hi @pepkin88 cdnjs adds versions based on git tags, so on cdnjs.com v1.5.0 seems v1.4.0. Maybe you can open an issue there to ask can they fix the tag. cc @PeterDaveHello

pepkin88:
I created a pull request for this issue, I thought it might me easier to update the library from a PR.

pepkin88:
There's a new version of LiveScript, 1.6.0, correctly loaded onto CDNJS. Although 1.5.0 will remain untouched, using 1.6.0 is preferable. Therefore I'll close this issue.

PR #12461, opened 2018-01-25
Fix LiveScript v1.5.0

pepkin88:
Updated to a correct version, was 1.4.0, is 1.5.0. Related to this issue: https://github.com/cdnjs/cdnjs/issues/12459.

I thought that merging this PR might be easier to do than triggering the auto-updater to redownload the library.


sufuf3:
@pepkin88 Hi, thank you for the PR.
However, CDNJS provide all the files from upstream, this means the files need to as the same with upstream. We can make sure every file are the same with the original files by hash checksum.

I know that you want to fix some problem through revised this version files and sent this PR.
But we hope the files which were hosted in CDNJS can be trusted.

So, the best way is to discuss the git version tag(like v1.5.1) with the author, then CDNJS can host the new version.

What do you think?

Many thanks.

pepkin88:
There's a new version of LiveScript, 1.6.0, correctly loaded onto CDNJS. Although 1.5.0 will remain untouched, using 1.6.0 is preferable. Therefore I'll close this issue.

sufuf3:
Thanks

PR #12853, opened 2018-07-01
[New] Update qunit auto-update and homepage

### LINKED ISSUE #12849: qunit no longer updating
cdnjs last qunit vesion is 2.4.1 which is when the npm package was renamed.
latest is 2.6.1

http://qunitjs.com/
> Prior to version 2.4.1, QUnit was published under the name qunitjs on NPM. If you wish to install an older version of QUnit on Node, you will want to use the qunitjs package. The qunit package prior to 2.4.1 is an alternative CLI that is now published as node-qunit.

looks like package.json just needs to be updated that the npmName should be `qunit` instead of `qunitjs`
https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/qunit/package.json#L22

also source is in /src/*

sashberd:
The npm name was update due to it was changed from qunitjs to qunit
According to: https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/qunit/package.json#L22
close #12849

Pull request for issue: #12849
# Git commit checklist

### DIFF
--- a/ajax/libs/qunit/package.json
+++ b/ajax/libs/qunit/package.json
@@ -3,7 +3,7 @@
   "filename": "qunit.min.js",
   "version": "2.4.1",
   "description": "An easy-to-use JavaScript Unit Testing framework.",
-  "homepage": "http://qunitjs.com/",
+  "homepage": "https://qunitjs.com/",
   "keywords": [
     "framework",
     "toolkit",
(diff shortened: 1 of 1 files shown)


sashberd:
@extend1994 The PR was updated. FYR

Issue #12813, opened 2018-06-07
[author][Request] Add datalist-polyfill


mfranzke:
**Library name:** datalist-polyfill
**Git repository url:** https://github.com/mfranzke/datalist-polyfill/
**npm package name or url (if there is one):** https://www.npmjs.com/package/datalist-polyfill
**License (List them all if it's multiple):** MIT
**Wanna say something? Leave message here:**
The downloads currently vary a little bit - yarn reported 11k the latest, whereas they are currently down to 6k: https://yarnpkg.com/en/package/datalist-polyfill

PeterDaveHello:
@mfranzke Send a pull request, please.

PR #12871, opened 2018-07-08
[author][new] Add datalist-polyfill w/ npm auto-update via single package.json

mfranzke:
Pull request for issue: #12813 
Related issue(s): # #

Note that if you are using a distribution purpose repository/package, please also provide the url and other related info like popularity of the source code repo/package.

# Profile of the lib
 * Git repository (required): https://github.com/mfranzke/datalist-polyfill/
 * NPM package url (optional): https://www.npmjs.com/package/datalist-polyfill
 * License and its reference: MIT
 * GitHub popularity (required):
   - Count of watchers: 8
   - Count of stars: 65
   - Count of forks: 24
   - weekly Downloads: 1795
   - weekly Downloads at its highest so far: 3618
   - total Downloads: 73k
   - jsDelivr hits/day: 70-80k
 * NPM downloads:
The downloads currently vary a little bit - yarn reported 11k the latest: https://yarnpkg.com/en/package/datalist-polyfill

# Essential checklist
 * [x] I'm the author of this library
   * [x] I would like to add link to the page of this library on CDNJS on website / readme
 * [x] This lib was not found on cdnjs repo
 * [x] Doesn't already exist / has duplicate issue and PR
 * [x] The lib has notable popularity
   * [ ] More than 200 [Stars / Watchers / Forks] on [GitHub / Bitbucket]
   * [x] More than 800 downloads stats per month on npm registry
 * [x] Project has public repository on famous online hosting platform (or been hosted on npm)

# Auto-update checklist
 * [x] Has valid tags for each versions (for git auto-update)
 * [x] Auto-update setup

# Git commit checklist

### DIFF
--- a/ajax/libs/datalist-polyfill/package.json
+++ b/ajax/libs/datalist-polyfill/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "datalist-polyfill",
+  "description": "A minimal and dependency-free vanilla JavaScript datalist polyfill. Supports all standard's functionality as well as mimics other browsers behavior.",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/mfranzke/datalist-polyfill.git"
+  },
+  "author": "Maximilian Franzke",
+  "license": "MIT",
+  "filename": "datalist-polyfill.min.js",
+  "keywords": [
+    "datalist",
+    "polyfill",
+    "javascript",
+    "html",
+    "autosuggest",
+    "autosuggester",
+    "suggest",
+    "suggester",
+    "select"
+  ],
+  "npmName": "datalist-polyfill",
+  "npmFileMap": [
+    {
+      "basePath": "",
+      "files": [
+        "datalist-polyfill*.+(js|map)"
+      ]
+    }
+  ]
+}


ajax/libs/datalist-polyfill/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "javascript",
+    "html",
+    "autosuggest",
+    "autosuggester",
+    "suggest",
+    "suggester",
+    "select"
+  ],
+  "npmName": "datalist-polyfill",
+  "npmFileMap": [
```
mfranzke: I do hope that these are filled out correctly - it's mainly the two JS files that would need to get hosted on a CDN for providing those to the users. And those are stored in the root of the project right away.

sashberd: Why do you use npm? On Git number of releases more than on npm.

ajax/libs/datalist-polyfill/package.json
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "basePath": "",
+      "files": [
+        "*.js"
```
sashberd: better to use `datalist-polyfill*`
Also in old versions like in 1.0.0 there are also css files that should be hosted

ajax/libs/datalist-polyfill/package.json
mfranzke: It‘s mainly about two of the earliest releases, that included both a typo and adding the package.json for npm usage - those wouldn‘t be necessary to reference any more, as they are way behind regarding functionality and bugfixes.

And as you seem to favor npm over github I‘ve than chosen that one to inlcude for auto-update.

ajax/libs/datalist-polyfill/package.json
mfranzke: Valid point ! But similar to the above: I‘ve deprecated the CSS dependency in the early days of the project, so this version wouldn‘t be necessary any more for reference. Additionally it shouldn‘t necessarly be referenced at all, but included in their project ecosystem directly, as it‘s content is too small, so it’s „cheaper“ in relation to the savings of an extra HTTP request (even from a CDN).

mfranzke:
@sashberd, thanks a lot for your valuable feedback! I‘ve replied to both of your comments and would like to suggest to leave the file as provided.

sashberd:
@extend1994 @sufuf3 See the authors remarks above my review. At your discretion.

ajax/libs/datalist-polyfill/package.json
mfranzke: Hi @extend1994,
thanks a lot for your feedback - and totally agreed, your mentioned change makes perfectly sense to me. I've updated the PR.

Issue #12877, opened 2018-07-09
[Request] Update Framework7


nolimits4web:
Framework7 us already v3.0.0 https://github.com/framework7io/framework7
cdnjs version is still 2.3.1

Also home page is changed a while ago:
http://idangero.us/framework7 -> http://framework7.io

Repository is changed too: 
https://github.com/nolimits4web/Framework7.git -> https://github.com/framework7io/framework7

Also would be good to have `framework7-vue` and `framework7-react` packages in addition as they sit in same repo

sashberd:
@nolimits4web I got your answer, as for vue and react please open separate issue for each one of the libs and provide your answer to my [question](https://github.com/cdnjs/cdnjs/pull/12878) as reference
Thanks

PR #12878, opened 2018-07-10
[New] Update framework7 auto-update and meta data

sashberd:
Update homepage property
Update repository property
Update auto update config
cc #12877

Pull request for issue: #12877 
Related issue(s): # #

# Git commit checklist

### DIFF
--- a/ajax/libs/framework7/package.json
+++ b/ajax/libs/framework7/package.json
@@ -33,13 +33,12 @@
   "homepage": "http://www.idangero.us/framework7/",
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/nolimits4web/Framework7.git",
+    "target": "git://github.com/framework7io/framework7.git",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": "packages/core",
         "files": [
-          "**/framework7!(.module|.modular|.esm|.less)*",
-          "img/**/*.+(png|svg)"
+          "**/framework7!(.module|.modular|.esm|.less)*"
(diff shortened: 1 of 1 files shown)


sashberd:
@extend1994 The PR was updated. FYR

ajax/libs/framework7/package.json
```diff
@@ -30,16 +30,15 @@
...
-  "homepage": "http://www.idangero.us/framework7/",
+  "homepage": "https://framework7.io",
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/nolimits4web/Framework7.git",
+    "target": "git://github.com/framework7io/framework7.git",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": "packages/core",
```
sufuf3: Not sure that should we also host `framework7-react.js`, `framework7-react.min.js`, `framework7-react.min.js.map` ...etc from https://github.com/framework7io/framework7/tree/master/packages/react.
As I know, React is the front-end framework, also here is an example for framework7-react: 
https://github.com/framework7io/framework7-react-static-template/blob/master/index.html#L18
Also, I think [vue](https://github.com/framework7io/framework7/tree/master/packages/vue) can be hosted, too. https://github.com/framework7io/framework7-template-vue-simple/blob/master/www/index.html#L339

What do you think?
Thank you.

sashberd: Lets ask authors opinion, @nolimits4web Would you like to answer @sufuf3 question?

Issue #12687, opened 2018-04-09
[Request] Add lottie-web


bodymovin:
Library name: lottie.js
Library description: A no depency JS library to render lottie.json files
Git repository url: https://github.com/airbnb/lottie-web
npm package name or url (if there is one): https://www.npmjs.com/package/lottie-web
License (List them all if it's multiple): MIT License
Official homepage: http://airbnb.io/lottie  

Wanna say something? This library replaces the bodymovin cdn files here:
https://cdnjs.com/libraries/bodymovin

MattIPv4:
Available on cdnjs under the old package name: https://cdnjs.com/libraries/bodymovin

ljharb:
@MattIPv4 the old package name is deprecated; this should be reopened, and it should be added under the new package name.

MattIPv4:
We don't currently have any logic in place to redirect users etc. when names get changed. As this still has usage under the old name, I guess the best option would be to add it under the new name and have both auto-update.

ljharb:
Right; it's not a redirection. A new name is a new package, and both packages should exist and match the published versions for them.

MattIPv4:
`bodymovin` is also set to update from `lottie-web`, so these are identical.

ljharb:
That just means two different packages are coincidentally identical :-)

PR #12905, opened 2018-08-06
Update bodymovin to the new lottie-web repository with v5.0.0 and filename change

DominikTo:
There is an request (#12687) as well as a PR to add lottie-web as a new lib (#12815), but since this library already exists on cdnjs (with it's old name bodymovin), maybe the right way to move forward is to rename/update the existing library (that stopped autoupdating once it was renamed in v5.0.0, see [this comment](
https://github.com/cdnjs/cdnjs/pull/12815#issuecomment-410786255)), instead of adding a new one.

Closing initial PR (#12904) and resubmitting, to make CI happy.

/cc @bodymovin


PeterDaveHello:
@sashberd you are correct, so we usually just update auto-update config without changing the name on CDNJS.

DominikTo:
Not sure why CI is still complaining about `/ajax/libs/lottie-web/` after undoing the renaming (bodymovin > lottie-web) - that directory doesn't exist.

/cc @PeterDaveHello

PeterDaveHello:
@DominikTo because it's still in the diff/change, the commits need to be squashed into a single one to avoid that ;)

DominikTo:
@sashberd The Problem is, that under the old lib name only versions up to 4.1.3 are available, so everybody who embedded the library via CDNJS under it's old name, won't see newer versions.

sashberd:
@DominikTo You said, what I said, but in another words. Lets finish  #12815 and leave this old lib as is. Then old and new lib will supported

DominikTo:
I'd still propose having "one lib" (there is no old lib and new lib; it just got renamed) that has _all_ version history, instead of one under the old name up to v4.1.3 and then a new one starting at v5.0.0 (that's also what @PeterDaveHello seemed to suggest), but: Whatever gets the latest version of this lib on CDNJS.

Feel free to close this PR, if this is not the way forward.

DominikTo:
B.t.w.: No idea why jsonlint is complaining here: 
https://ci.cdnjs.com/cdnjs/cdnjs/22558

ajax/libs/bodymovin/package.json
```diff
@@ -15,13 +15,13 @@
     "export"
   ],
   "license": "MIT",
-  "homepage": "https://github.com/bodymovin/bodymovin/wiki",
+  "homepage": "https://github.com/airbnb/lottie-web/blob/master/LICENSE.md",
```
sashberd: The homepage should be https://airbnb.io/lottie/

sashberd: The homepage should be https://airbnb.io/lottie/

DominikTo: Updated.

DominikTo:
Thanks for your feedback @PeterDaveHello.

I'd be totally fine with just getting this merged, to get versions >=5.0.0 under the existing/old lib name on CDNJS.

Maybe for the future (not related to this issue) there should be some general approach to rename libs (and potentially forward/alias from old name).

bodymovin:
Hi, lottie dev here.
I'm fine too using the old player name.
If people get confused we can revisit this decision in the future.

PeterDaveHello:
@bodymovin thank you so much

sashberd:
@PeterDaveHello Thus #12815 has to be closed?

DominikTo:
Unless there are any remaining issues, It would be great to merge this, before @PeterBot complains that the branch is too old and has to be rebased. /cc @sashberd, @PeterDaveHello.

DominikTo:
Thanks, @sashberd!

DominikTo:
Should this one also get tagged as in progress?
/cc @sashberd, @PeterDaveHello

bodymovin:
Hi, is there anything missing that I can help with?

PR #13088, opened 2018-11-09
Update split.js to new repo

nathancahill:


PeterDaveHello:
@extend1994 great :+1: BTW, I think you can directly merge this kind of PR next time, though the bot ignores the structure changes on map file, which means v1.5.9 looks the same as v1.5.7, can be automatically added/updated!

extend1994:
@PeterDaveHello Got it!

Issue #13112, opened 2018-12-03
Update svg.js


Fuzzyma:
**Library name:** svg.js
**Git repository url:** https://github.com/svgdotjs/svg.js
**npm package**: https://www.npmjs.com/package/@svgdotjs/svg.js
**License:** [MIT](https://github.com/svgdotjs/svg.js/blob/master/LICENSE.txt)
**Homepage**: svgjs.com

We do not upload distribution files to github anymore from version 3 on. So you need to use the npm package instead or use the zip from the release page

PeterDaveHello:
@Fuzzyma feel free to send a pull request :)

Fuzzyma:
@PeterDaveHello Thanks for the fast response. What do I need to change so that npm is used instead of github as endpoint?

PR #13113, opened 2018-12-03
Update svg.js to use npm auto-update

Fuzzyma:
Pull request for issue: #13112 

**This libraray is already available on cdnjs. So all checks are fullfilled.**

As requested, I opened a PR to change the svg.js endpoint to npm.


ajax/libs/svg.js/package.json
```diff
@@ -10,22 +10,19 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/svgdotjs/svg.js/"
   },
+  "npmName": "@svgdotjs/svg.js",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "*"
```
extend1994: Hi, @Fuzzyma I'm wondering `svg.node.js` is for node environment, is that so?

Fuzzyma: Yes thats true. Shall I exclude it? If so, is `!(.*)node(.*)` sufficiant?

extend1994: Maybe `**/!(*.node.*)`? It grabs all files except for the ones with the filename`.node.` recursively.

Fuzzyma: Alright, will use that. Thank you!

Fuzzyma:
To rebase I would need to pull at least the spare repo. However pulling 2Gigs of files is not possible for me.
I did all changes directly inline on github so I  dont need to pull the repo. Can you rebase for me? (since I only changed one folder alone, there should be no rebase required anyway)

PR #13151, opened 2018-12-15
Update git url for SortableJS

christiansk:
It seems the github repo for this library was moved out of the original developer's github account to its own project account.
- Old repo: https://github.com/RubaXa/Sortable
- New repo: https://github.com/SortableJS/Sortable
Also seems the latest version is v1.7.0.

### DIFF
--- a/ajax/libs/Sortable/package.json
+++ b/ajax/libs/Sortable/package.json
@@ -1,7 +1,6 @@
 {
   "name": "Sortable",
   "title": "Sortable",
-  "version": "1.7.0",
   "filename": "Sortable.min.js",
   "homepage": "http://rubaxa.github.io/Sortable/",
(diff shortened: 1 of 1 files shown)


ajax/libs/Sortable/package.json
```diff
@@ -1,13 +1,12 @@
 {
   "name": "Sortable",
   "title": "Sortable",
-  "version": "1.6.0",
```
PeterDaveHello: Please do not manually touch the version info here.

christiansk: Sorry about that; still learning how this works.

christiansk:
@PeterDaveHello I returned the version info to what it was previously. I understand the latest version will always be pulled from the project's git repom is that correct?
Is there anything else I should do to get this merged?
Thanks!

christiansk:
Closing this since apparently it's not needed. I think due to the Github redirection to the new repository new releases are already being fetched from there.

PR #13154, opened 2018-12-15
Update jquery.blockUI license info in package.json

ashokjayaprakash:
Reference: https://github.com/malsup/blockui#copyright-and-license
CC: #11931

### DIFF
--- a/ajax/libs/jquery.blockUI/package.json
+++ b/ajax/libs/jquery.blockUI/package.json
@@ -16,10 +16,7 @@
   },
   "licenses": [
     "MIT",
-    {
-      "type": "GPL",
-      "url": "http://malsup.github.com/gpl-license-v2.txt"
-    }
+    "GPL-2.0"
   ],
   "repository": {
     "type": "git",


ajax/libs/jquery.blockUI/package.json
```diff
@@ -14,13 +14,7 @@
...
     "url": "http://jquery.malsup.com"
   },
-  "licenses": [
-    "MIT",
-    {
-      "type": "GPL",
-      "url": "http://malsup.github.com/gpl-license-v2.txt"
-    }
-  ],
+  "license": "(MIT OR GPL-2.0-or-later)",
```
extend1994: Let's use
```js
  "licenses": [
    "MIT",
    "GPL-2.0"
  ]
```

ashokjayaprakash:
@extend1994 : Updated as per your review comments. As per [npm latest documentation](https://docs.npmjs.com/files/package.json#license) , it is suggests to use SPDX license expression syntax version 2.0. Just wanted to understand the reason behind it.

extend1994:
@ashokjayaprakash `package.json` here is only for cdnjs to parse info in cdnjs.com or grab library versions, etc. Therefore, it may be different from the npm suggestion.
`license` here should be the license format defined in https://github.com/cdnjs/cdnjs/blob/master/tools/license-list.json which is SPDX v2.5.

PR #13155, opened 2018-12-15
Update jquery.collapsible license info in package.json

ashokjayaprakash:
Reference: https://github.com/juven14/Collapsible/blob/master/LICENSE 
CC: #11931

### DIFF
--- a/ajax/libs/jquery.collapsible/package.json
+++ b/ajax/libs/jquery.collapsible/package.json
@@ -13,10 +13,7 @@
     "div"
   ],
   "version": "1.2",
-  "license": {
-    "type": "BSD",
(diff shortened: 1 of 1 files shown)


ajax/libs/jquery.collapsible/package.json
```diff
@@ -13,10 +13,7 @@
     "div"
   ],
   "version": "1.2",
-  "license": {
-    "type": "BSD",
-    "url": "http://www.snyderplace.com/collapsible/license.txt"
-  },
+  "license": "BSD",
```
extend1994: It's actually `BSD-3-Clause`
> This license has also been called the "New BSD License" or "Modified BSD License"

See https://opensource.org/licenses/BSD-3-Clause for the explanation

PR #13365, opened 2019-05-18
[Update] webui-popover

MattIPv4:
Fix to an existing library.

The library makes reference to the img folder from assets inside the dist folder so it will be best for us to include both the dist & img folders.

### DIFF
--- a/ajax/libs/webui-popover/package.json
+++ b/ajax/libs/webui-popover/package.json
@@ -1,5 +1,5 @@
 {
-  "filename": "jquery.webui-popover.min.js",
+  "filename": "dist/jquery.webui-popover.min.js",
   "name": "webui-popover",
   "description": "A enhancement popover plugin for Bootstrap ,but you can use it stand-alone without Bootstrap!",
   "author": "Sandy Duan <sanddywalker@gmail.com>",
@@ -22,7 +22,7 @@
         "basePath": "",
         "files": [
           "dist/*",
-          "img/*
+          "img/*"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


MattIPv4:
@PeterDaveHello What is the best way to fix this? All versions we have are missing the required img folder, but a restructure would mean all existing usage breaks?

ajax/libs/webui-popover/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "filename": "jquery.webui-popover.min.js",
+  "filename": "dist/jquery.webui-popover.min.js",
```
PeterDaveHello: Should not touch this line

PeterDaveHello:
I think we should let the upstream fix it as it's beyond our topics?

MattIPv4:
@PeterDaveHello This isn't really an issue with upstream though, we just aren't using the correct structure from the upstream?

PeterDaveHello:
@MattIPv4 I mean in convention or common cases, dist is for distribution, reference a file not being distributed doesn't make sense.

MattIPv4:
@PeterDaveHello Yeah, fair enough, I was just going for the approach of resolving this ourselves instead of waiting on upstream to resolve it.

Should we just an open an issue with the upstream provider instead?

PeterDaveHello:
@MattIPv4 yes I prefer that, and in most cases, upstream would also like to know the issues, unless reaching out doesn't work at all, I prefer not to fix or add workaround by our side, thanks.

Issue #11123, opened 2017-04-23
Migrate to sha384/512 on SRI hash calculation


PeterDaveHello:
Issues:
1. We'll need a powerful and individual platform to focus on re-calcuating all the hashes over all the 950k js & css files?
2. Should the origin sha256 hashes be preserved? Is the answer is yes, how do we deal with it? (How to display on the UI? How to save both sha256 and sha384 hashes?)
3. How urgent we **need** it?
4. How's the impact on performance on both hash calculation and website hosting? Especially the calculation time and memory consumption part.

MattIPv4:
Bringing @drewfreyling's comment from #13384:

I'm in favour of just offering up the SHA-512 values since it is:

 - Less bytes in the browser
 - All browsers (currently) must support 256, 384 and 512 (https://w3c.github.io/webappsec-subresource-integrity/#hash-functions)

If at some point we get fragmented browser support than I think providing multiple hashes makes sense.

myshov:
@PeterDaveHello @MattIPv4 
Hello guys
I suggest to prioritize this issue because SHA-256 and SHA-512 are susceptible to length extension attacks. I suggest to use SHA-384 because it has no this flaw https://blog.skullsecurity.org/2012/everything-you-need-to-know-about-hash-length-extension-attacks

MattIPv4:
cc @zackbloom @xtuc

MattIPv4:
We should keep this open as we are still looking at implementing this.

PeterDaveHello:
Feel free to open a new issue for it(or maybe just use the opened pull request for discussion), as I was removed from the organization without any discussion and respect, just like some other members(I know it's not by you and you're not the one who's in charge for that), I'm now closing ancient issues I'm not working on so that I don't need to receive unnecessary notifications, thanks.

ExE-Boss:
Did you know that you can simply unsubscribe from issues you no longer wish to receive notifications for?

<table><td><img src="https://user-images.githubusercontent.com/3889017/75113480-a55d1480-564e-11ea-96c0-b8c07afa6b7e.png"/></table>

It’s what that button in the sidebar is there for.

MattIPv4:
Noting this down: the current SRI generation happens in https://github.com/cdnjs/tools/blob/master/packages/packages.go#L86 & https://github.com/cdnjs/tools/blob/master/openssl/sri.go

klausenbusk:
[...]

Issue #13384, opened 2019-06-18
SRI SHA 384/512


MattIPv4:
Following a [user suggestion on the community forum](https://cdnjs.discourse.group/t/sri-sha-384-512-support/79), I believe that we should implement SHA-384 and SHA-512 alongside the existing SHA-256 in the SRI values that we provide to users and use.

```
<!-- Current format -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.4.1/jquery.min.js" integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"></script>
 
<!-- Suggested format -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.4.1/jquery.min.js" integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo= sha384-vk5WoKIaW/vJyUAd9n/wmopsmNhiy+L2Z+SBxGYnUkunIxVxAv/UtMOhba/xskxh sha512-bnIvzh6FU75ZKxp0GXLH9bewza/OIw6dLVh9ICg0gogclmYGguQJWl8U30WpbsGTqbIiAwxTsbe76DErLq5EDQ==" crossorigin="anonymous"></script>
```

MattIPv4:
Relevant code looks to be https://github.com/cdnjs/cdnjs/blob/master/build/packages.json.js#L205-L237

drewfreyling:
I'm in favour of just offering up the SHA-512 values since it is:

- Less bytes in the browser
- All browsers (currently) must be support 256, 384 and 512 (https://w3c.github.io/webappsec-subresource-integrity/#hash-functions)

If at some point we get fragmented browser support than I think providing multiple hashes makes sense.

PeterDaveHello:
dup https://github.com/cdnjs/cdnjs/issues/11123

MattIPv4:
@PeterDaveHello Completely missed that, I had only searched for 512.

PeterDaveHello:
Let's just move to #11123, BTW, I don't think it's a beginner issue.

PR #13386, opened 2019-06-19
Add sha384 and sha512 to SRI

danielfain:
This is for issue #11123

## Git commit checklist

### DIFF
--- a/build/packages.json.js
+++ b/build/packages.json.js
@@ -213,12 +213,23 @@ glob('ajax/libs/*/package.json', function (error, matches) {
           switch (fileType) {
             case 'js':
             case 'css':
-              var genSRI = 'cat "' + filespec + '" | ' +
+              var gen256SRI = 'cat "' + filespec + '" | ' +
                 'openssl dgst -sha256 -binary | ' +
                 'openssl enc -base64 -A';
-              var fileSRI = execSync(genSRI).toString();
-              if (fileSRI !== '') {
-                libSri[temp.files[i]] = 'sha256-' + fileSRI;
+              var gen384SRI = 'cat "' + filespec + '" | ' +
+                'openssl dgst -sha384 -binary | ' +
+                'openssl enc -base64 -A';
+              var gen512SRI = 'cat "' + filespec + '" | ' +
+                'openssl dgst -sha512 -binary | ' +
+                'openssl enc -base64 -A';
+              var file256SRI = execSync(gen256SRI).toString();
+              var file384SRI = execSync(gen384SRI).toString();
+              var file512SRI = execSync(gen512SRI).toString();            
+
+              if (file256SRI !== '' && file384SRI !== '' && file512SRI !== '') {
+                libSri[temp.files[i]] = 'sha256-' + file256SRI + 
+                  ' sha384-' + file384SRI + 
+                  ' sha512-' + file512SRI;
               }
 
               break;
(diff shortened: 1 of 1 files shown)


MattIPv4:
Hey, @danielfain thanks for jumping on this super quick!

Per the suggestion of @drewfreyling in #13384 I think it might be better to just generate the SHA-512. However, I think it'd be great to leave the existing code you've written in place for 256/384 in the event we decide to ever use them, just comment it out for now :)

danielfain:
Hey, @MattIPv4 this is my first time contributing to an open source project so I'm not quite sure the best way to tweak it. Should I just make another commit on my branch with the aforementioned fixes or should I rebase both commits into 1?

Thanks!

MattIPv4:
@danielfain If you're happy with rebasing, a single commit would be nicer :)

build/packages.json.js
```diff
@@ -213,14 +213,32 @@ glob('ajax/libs/*/package.json', function (error, matches) {
...
+              if (file256SRI !== '' && file384SRI !== '' && file512SRI !== '') {
+                libSri[temp.files[i]] = 'sha256-' + file256SRI + 
+                  ' sha384-' + file384SRI + 
+                  ' sha512-' + file512SRI;
+              }
+              */
+              if (file512SRI !== '') {
+                libSri[temp.files[i]] = 'sha512-' + file512SRI;
               }
-
```
PeterDaveHello: this line change is not related

MattIPv4:
@xtuc What are your thoughts on this?

xtuc:
That's nice. The right place is now in https://github.com/cdnjs/tools/blob/master/openssl/sri.go#L10.

Also I'm wondering if this should apply to the existing assets on the cdn, because it will take a fair amour of time to generate.

My understanding is some non-browsers could rely on the hash via the API as well, shouldn't we support both?

MattIPv4:
Yeah, that was the main issue when this was originally suggested - do we just jump to 512? Do we provide 256, 384 & 512 in a single string? Do we default to 512 and provide 256/384 via additional SRI attributes in the API? Do we generate the new hashes for all lib files? Do we only generate for new files once implemented?

MattIPv4:
To me, either of these two makes sense:

Jump to 512 and only generate it going forward. Old assets will still just use 256 and anything new will just use 512.

Providing 512 by default with 256 & 384 available as additional attrs. And, we generate the hashes for all files on cdnjs, so that the API always has correctly populated SRI data.

rarkins:
Hi, we are mostly done implementing cdnjs updating in [Renovate Bot](https://github.com/renovatebot/renovate). The problem I'm seeing is that we are "downgrading" the sha strength for people who have `sha384` or `sha512`. Although the theory behind "only sha512" seems fine to me, I'd still prefer that all of the hashes are made available in the API so that we can match with the one they currently have and cause least "friction'.

Here's an example of a PR today when we use the SRI from the cdnjs API:

image

MattIPv4:
SRI generation is no longer handled through this script. Closing.
We can look at supporting stronger SRI hashes through the new scripting.

rarkins:
@MattIPv4 do you have any issue or link I can follow?

MattIPv4:
I would continue tracking https://github.com/cdnjs/cdnjs/issues/11123, we'll post updates there as we go.

PR #13400, opened 2019-07-08
Switches 'raty' package to npm update

ecabuk:
git auto update is enabled for this package (raty) but it looks like author released a new version on the npm but it doesn't tagged it on GitHub and because of that CDNJS does not include the new version 2.9.0

According to your docs `npm` is the preferred way to update packages, so I thought this PR would work in both ways.

https://github.com/cdnjs/cdnjs/blob/3df219e22a51443dfa76c9ea535b672df36f1f2b/ajax/libs/raty/package.json#L25-L38

Related Links:

https://www.npmjs.com/package/raty-js
https://github.com/wbotelhos/raty


ajax/libs/raty/package.json
```diff
@@ -23,17 +23,14 @@
...
-      }
-    ]
-  }
+  "npmName": "raty-js",
+  "npmFileMap": [
+    {
+      "basePath": "lib",
+      "files": [
+        "**/?(jquery.)raty*", 
+        "images/*.png"
```
MattIPv4:
```suggestion
        "**/*", 
```
The fonts might also be needed, so a glob for everything in `lib` seems best.

ajax/libs/raty/package.json
```diff
@@ -23,17 +23,13 @@
...
-        ]
-      }
-    ]
-  }
+  "npmName": "raty-js",
+  "npmFileMap": [
+    {
+      "basePath": "lib",
+      "files": [
+        "**/*",
```
MattIPv4:
```suggestion
        "**/*"
```
Rm trailing comma

PR #13442, opened 2019-08-30
Update mathjax auto-update config for mathjax v3

dpvc:
This PR is in preparation for the version 3.0.0 release that is planned for this weekend.  It adds the `es5` direcotory that is new in version 3.0.0.  It also updates the version number to the current version (2.7.6), though after tomorrow, the current version will be 3.0.0.

You won't see the `es5` directory in the MathJax repository until the release is made this weekend.

## Auto-update checklist

* [x] Has valid tags for each versions (for git auto-update)
* [x] Auto-update setup

## Git commit checklist

* [ ] Close corresponding issue in commit message
* [ ] Mention related issue(s), people in commit message, comment.


dpvc:
OK, I rebased the PR.  Please let me know if there is anything else needed.  I'm not sure how to let PeterBot know that the change has been made.  Do I need to make a new PR?

PeterDaveHello:
@dpvc I'll take a look.

dpvc:
@PeterDaveHello, can you give me any idea when this PR might be reviewed?  We are holding up the v3 release until this has been merged, so that the 3.0.0 release will appear on cdnjs properly.  All our documentation and examples point to cdnjs, and we don't want to make the release if the cdnjs version won't be correct.

PeterDaveHello:
@dpvc hmmm, I'll try to make some spare time on this weekend, hopefully it's good for you, I wish I can get it done ASAP but unfortunately we open source project maintainers still need to make money to keep alive, so won't respond too quick sometimes.

dpvc:
@PeterDaveHello, OK, thank you.  I didn't mean to be an annoyance.  I had hoped that since the change is really only one line in the package file (plus the version number), it might not be too burdensome to handle.

dpvc:
PS, I am also an ope-source project maintainer, so I understand your situation.  Thanks for your work on cdnjs, which is a great service.

PeterDaveHello:
@dpvc no problem

ajax/libs/mathjax/package.json
```diff
@@ -21,6 +21,7 @@
       {
         "basePath": "",
         "files": [
+          "es5/**/!(node-main.js)",
```
PeterDaveHello: Looks like the other patterns will not be needed? Which file would be the main file then?

dpvc: We will be continuing to support version 2 for some time (version 3 is a major rewrite and has breaking changes), and expect to be issuing updates to version 2.  These will need to old file structure, which is why I didn't remove them.  That way both version 3 and version 2 will update properly.

Version 3 doesn't really have a main file, but I suppose `es5/startup.js` would be the best alternative for that.

zorkow:
Is there any news on this? When do you expect to merge v3?

dpvc:
@MattIPv4, thank you for reviewing this pull request.

> would it make sense to update the `basePath` here to now be `es5`...?

If I understand that correctly, that would mean that the top level of the CDN would be the contents of the `es5` directory, and `es5/` would no longer be in the URL of the files served from cdnjs?  If so, then this would not be ideal for us.  We are planning to have an `es6` directory in the future, and would not want to hierarchy to change when that is included (that is why there is an `es5` directory now).  Also, as mentioned tin a comment above, we are still making updates to the 2._x_ versions, and that would prevent them from being updated properly.  That is why the other patterns are still in place.

> the `filename` property in the `package.json` file will need updating

I have changed it to `es5/startup.js`.

Thank you agin for your help.

MattIPv4:
Ah okay, makes sense. I'm not sure if the auto-update bot will pull in new 2.x versions once it discovers a 3.x version, we'll have to see. I agree with keeping the file map as is though, based on that.

dpvc:
Thanks for the information, and the caution about future v2 versions.  We will be making a v2.x update in the next month, I expect, so will soon know the result.  Versions 3.0.0 and 3.0.1 have been released since this PR was first posted last August.  Will it pick those up automatically, or is there something else I should do for those?

MattIPv4:
The bot should just pick up the new versions as it hasn't done yet, likely due to the current auto-update config matching no files in those versions.

dpvc:
> The bot should just pick up the new versions

Great!

Thank you again for helping guide this PR through completion.

PR #13451, opened 2019-09-08
Fix zTree.v3

### LINKED ISSUE #13429: update ztree.v3 to 3.5.40
now have ztree 3.5.40.
but download this result is ztree 3.5.0.1.
i hope update ztree to 3.5.40
image

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/78791022-update-ztree-v3-to-3-5-40?utm_campaign=plugin&utm_content=tracker%2F32893&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F32893&utm_medium=issues&utm_source=github).
</bountysource-plugin>

PeterDaveHello:
Ref: https://github.com/cdnjs/cdnjs/issues/13429#issuecomment-529184862

This PR will fix #13429 and close #13447
    
cc cdnjs/autoupdate@b6076a7d0e7cfec1bf9885840c025ab8dcbc7a3d


PeterDaveHello:
Before this getting merged, we should at least announce this "fix" as it'll also break those websites referring this library with SRI enabled, and even better if we can get also have help from zTree to announce from their side, together.

@MattIPv4 can you help with this?

MattIPv4:
@PeterDaveHello Are we sure we want to be updating existing assets?

Imo, we should just fix this for new versions but leave the old bad versions so that we don't cause SRI issues and can retain the promise that assets on the CDN never get changed once published.

PeterDaveHello:
Actually I don't want to do that but looks like almost the whole zTree.v3 is wrong, if we can trace back to those website using this library by referrer header, maybe we can confirm if they enabled SRI and will be broken or not.

MattIPv4:
Sure thing, I'll reach out to Cloudflare and see if they can help!

MattIPv4:
@PeterDaveHello You should be cc'ed into the Cloudflare support ticket with the log dump. I don't want to post it publicly here as it contains some sites that are raw IP addresses :)
Overall, looks to be very few sites using this, so it might be safe to do. Your call :)

PeterDaveHello:
@MattIPv4 I'd take a look soon, thanks!

MattIPv4:
After some further discussion internally, we've decided it best that we don't interfere with existing published assets and invalid their SRIs.

PR #13542, opened 2019-11-05
Sync with latest git repos

xtuc:


xtuc:
Closing because I will split it in multiple commits instead.

PR #13543, opened 2019-11-05
Sync with git repos - part1

xtuc:


PeterDaveHello:
Thanks, not sure what I can do here right now, this looks huge to me to be reviewed, consider of the model how cdnjs works, once the assets are merged into `master` branch (I see it's `next` not `master` here) and hosted, the hash is calculated, anything wrong could be very hard to be fixed 🤔 I think we still have some discussions undergoing, maybe can wait a little bit to make sure everything is correct.

zackbloom:
Is there another way to get things up to date you can think of @PeterDaveHello?

PeterDaveHello:
@zackbloom as I mentioned earlier in the mail, the bot may be better in this case. Though it can't handle all the new versions, at least the added ones would usually be no mistake,

xtuc:
Note that I ran the autoupdater https://github.com/cdnjs/autoupdate

PeterDaveHello:
Forgot to update here, since I saw the huge PR here, and don't know more details yet, I just spend some time to bring the bot up again, so it can buy us more time to deal with all the things and details. The bot is basically use the autoupdater to check updates, but do more additional check to make sure the file/directory structure broken, I'll suggest to use autoupdate tool with manual or bot checking to make sure everything is correct.

PR #13604, opened 2019-11-06
Update form npm via bot

xtuc:


xtuc:
`web-minify-helper/minify.sh` wasn't applied

PR #13607, opened 2019-11-10
[author] Update to GSAP 3 (GreenSock)

jackdoyle:
There's a new directory structure to GreenSock's Github repo, so the files to distribute are in the /dist directory now (and we renamed the repo to simply "GSAP")

* [X] I'm the author of this library
* [X] The lib has notable popularity


jackdoyle:
@PeterBot Thanks for the prompt attention!

From what I can tell, the problem is that the files haven't been pulled in yet via the auto-update...but that was the point of the PR (to update the location so that things get pulled in properly). In other words:

> /ajax/libs/gsap/3.0.0/gsap.min.js does not exist but is referenced in package.json! 

...once the auto-update runs, that file WILL exist, right? Or do I need to manually add the files and create another PR? If so, the Github UI doesn't let me create a new directory (in https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/gsap/) so if I need to do everything with a full fork, this could get quite time-consuming/painful :) (but maybe I'm missing something). 

Help? 

CC @MattIPv4 (who has been helping via Twitter DM)

MattIPv4:
@jackdoyle Just DM'ed you about that, you'll need to include the files for the version this PR is bringing the package.json version up to. Auto-update will then pull in future versions.

MattIPv4:
> If so, the Github UI doesn't let me create a new directory

You can do it in this PR, if you go to create a new file using the GitHub UI, just type `directory/name.ext` as the name and it'll create `directory` for you. GitHub is magic.

jackdoyle:
I'll create a new pull request with everything shortly...

PR #13616, opened 2019-11-14
HTTPS link to https://lunrjs.com

coliff:


XhmikosR:
Shouldn't this be made upstream to the lunr.js package?

coliff:
It is updated there already... not sure if it gets updated here automatically upon the next release though.

XhmikosR:
OK so this PR is moot then :) When the package is updated here, it'll be fixed.

coliff:
cool. no problem, I'll close this PR then.

PR #13617, opened 2019-11-14
Docs: Update readme

MattIPv4:
- Uses the cdnjs “brand header” in the readme doc
 - Removes the broken wafflebot throughout graph
 - Updates intro section with more info on our other repos


MattIPv4:
As these docs will soon be replaced with new docs that can be found in https://github.com/cdnjs/packages-migration/tree/master/new-docs during the repository migration, I'm not sure there is much point in making this change here.

PR #13659, opened 2020-01-15
[new] [author] add sanctuary

davidchambers:
:raising_hand_woman: _I would appreciate help with this pull request as it is my first contribution to this project. :)_

## Profile of the lib

* Git repository (required): <https://github.com/sanctuary-js/sanctuary>
* Official website (optional, not the repository): <https://sanctuary.js.org/>
* NPM package url (optional): <https://www.npmjs.com/package/sanctuary>
* License and its reference: <https://github.com/sanctuary-js/sanctuary/blob/master/LICENSE> (MIT)
* GitHub / Bitbucket popularity (required):
  * Count of watchers: 58
  * Count of stars: 2300
  * Count of forks: 94
* NPM download stats (optional):
  * Downloads in the last day:
  * Downloads in the last week:
  * Downloads in the last month:

## Essential checklist

* [x] I'm the author of this library
  * [ ] I would like to add link to the page of this library on CDNJS on website / readme
* [x] This lib was not found on cdnjs repo
* [x] Doesn't already exist / has duplicate issue and PR
* [x] The lib has notable popularity
  * [x] More than 200 [Stars / Watchers / Forks] on [GitHub / Bitbucket]
  * [x] More than 800 downloads stats per month on npm registry
* [x] Project has public repository on famous online hosting platform (or been hosted on npm)

## Auto-update checklist

* [x] Has valid tags for each versions (for git auto-update)
* [x] Auto-update setup
* [ ] Auto-update target/source is valid.
* [ ] Auto-update filemap is correct.

## Git commit checklist

* [x] The first line of commit message is less then 50 chars; clean, clear and easy to understand.
* [x] The parent of the commit(s) in the PR is not older than 3 days.
* [x] Pull request is sent from a non-master branch with a meaningful name.
* [x] Separate unrelated changes into different commits.
* [x] Use rebase to squash/fixup dummy/unnecessary commits into only one commit.
* [x] Close corresponding issue in commit message
* [x] Mention related issue(s), people in commit message, comment.

PR #13665, opened 2020-01-17
remove outdated reference to automatic minification

davidchambers:
According to @MattIPv4 in <https://github.com/cdnjs/cdnjs/pull/13659#discussion_r367814197>, automatic minification is no longer enabled.


MattIPv4:
As these docs will soon be replaced with new docs that can be found in https://github.com/cdnjs/packages-migration/tree/master/new-docs during the repository migration, I'm not sure there is much point in making this change here.

MattIPv4:
Having spent some time recently working on the auto-update bot, it is believed the in-place minification is still active within the auto-update bot, so this change is no longer needed.

PR #13678, opened 2020-02-07
Create .travis.yml

xtuc:


MattIPv4:
If we can, IMO, it'd be nicer to run CI with Actions instead of external, now that GitHub offers integrated CI?

xtuc:
Travis doesn't seem happy with the size of the cdnjs repo... Do you think GitHub Actions would allow it? Ideally the CI would keep a copy of the entire repo instead of re-cloning for each run.

MattIPv4:
I'm not sure if anyone is ever happy with the size of the repo - with Actions though it won't automatically checkout the repo, so we could generate a sparse config based on the contents of the PR, checkout the repo with that and then run the linter?

xtuc:
That's a good idea

MattIPv4:
(I think our old CI did that, but I'm not sure though)

PR #13726, opened 2020-03-12
[update] Change swiper package.json npm auto-update

lizheming:
The latest swiper module version is `5.3.6`, while it still is `4.5.1` in cdnjs. I found swiper changed its structure at 5.x https://www.jsdelivr.com/package/npm/swiper?version=5.3.6, `dist/` file is disappeared. 

So I update its info to make auto-update work again, please take a look, thank you~


ajax/libs/Swiper/package.json
```diff
@@ -16,9 +16,9 @@
   "npmName": "swiper",
   "npmFileMap": [
     {
```
xtuc: not entirely sure but I think we should keep the `basePath`. 
```suggestion
    {
      "basePath": "",
```

lizheming: Emm.... althought it has a little weird, I have commited it please check again~

PR #13831, opened 2020-04-15
Update popper.js w/ npm auto-update

nkr0:
perhaps the change in package name from `popper.js` to `popper-core` demands a new folder in cdnjs, similar to the new package in npmjs.

### DIFF
--- a/ajax/libs/popper.js/package.json
+++ b/ajax/libs/popper.js/package.json
@@ -1,11 +1,11 @@
 {
   "name": "popper.js",
   "filename": "popper.min.js",
-  "version": "1.16.1",
+  "version": "2.3.3",
   "description": "A kickass library to manage your poppers",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/FezVrasta/popper.js.git"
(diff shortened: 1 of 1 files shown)


ajax/libs/popper.js/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "popper.js",
   "filename": "popper.min.js",
-  "version": "1.16.1",
+  "version": "2.3.3",
```
MattIPv4, line 4:
```suggestion
  "version": "1.16.1",
```
Please revert this change, the auto-update bot handles changing the version when it pulls inn new versioned files.

ajax/libs/popper.js/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "popper.js",
   "filename": "popper.min.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "umd/popper.min.js",
```
Whilst we're here, let's update this to reflect what's in the package.

PR #13965, opened 2020-05-25
autoupdate: switch from github to npm

alfredwassermann:
Since auto-update from github still does not work for JSXGraph we switch to npm.
Thanks for the good work!


alfredwassermann:
I tried to do that, but it is rejected because of:
    
    GitHub Actions/ robotcdnjs
    packages/j/jsxgraph.json#L1
    package download per month on npm is under 800

What can be done?

MattIPv4:
That's fine. The library is already on cdnjs so we can ignore that.

MattIPv4:
https://github.com/cdnjs/packages/pull/178

PR #13994, opened 2020-07-02
Cleanup

klausenbusk:
https://github.com/cdnjs/packages/pull/292#issuecomment-653103620
> Hey @klausenbusk, would you be interested in submitting a similar PR on cdnjs/cdnjs? We'd want the npmName/npmFileMap transformation & the pretty printing, but we don't need to remove version over there.

/cc @MattIPv4


MattIPv4:
@xtuc @tc80 does this look good to you?

Assuming it does, once we're ready to merge we can just open the PR again & merge right away (leaving it open will result in the bot auto-closing again).

xtuc:
I don't think we will be able to merge a PR given that the bot is pushing and another bot is closing. What if I run the commands on the repo and commit them?

MattIPv4:
We can either run the commands ourselves, or we could wait until @klausenbusk is around, then update the PR to match master (might be easier to essentially drop the current commit, pull latest master & rerun the fix commands?) and immediately reopen + merge?

Running the commands directly on master might be easiest to avoid any more conflicts?

klausenbusk:
> & the pretty printing

Does the bot "pretty print" the file? If not that should probably be fixed first.

MattIPv4:
The bot doesn't currently pretty-print, but the old one did. Pretty printing isn't technically needed here, I just opted to ask for it so that we could more easily look at the diff and check properties were transformed correctly.

If we run this directly on master instead of via a PR, we could just run the npmName transformation command only?

klausenbusk:
> If we run this directly on master instead of via a PR, we could just run the npmName transformation command only?

Correct, you probably want to run `jq` with `-c compact instead of pretty-printed output;`

MattIPv4:
@xtuc Do you just want to do that then, it seems like the simplest approach to get the transformation done to old packages here?

xtuc:
I ran the commands on the repo  https://github.com/cdnjs/cdnjs/commit/45751f30214fd573c7c360b74191cd520e52b91f and https://github.com/cdnjs/cdnjs/commit/f95097aaf400ab856baa45339630148af37e84d0

PR #13995, opened 2020-07-04
Fix ally.js

klausenbusk:
The version in `package.json` doesn't exist and isn't the latest version. `autoupdate` can't add `v1.3.1` as the files isn't stored the usual place (in the root).

Fix https://cdnjs.com/libraries/ally.js (ally.js version not found!)

/refs: https://github.com/cdnjs/packages/issues/302
/cc @MattIPv4


MattIPv4:
Manually adding files isn't something we support.

@xtuc could you directly commit to master from the bot just to update the version for `ally.js` to `1.4.1`?

klausenbusk:
> @xtuc could you directly commit to master from the bot just to update the version for `ally.js` to `1.4.1`?

We need both changes else `autoupdate` will try adding `v1.3.1` forever.

We also need to fix 3Dmol, it also can't add `v1.3.1` (the relevant files is stored in `release`, but in `v1.3.0` and `v1.3.2` the files is stored in `build` (?!))

MattIPv4:
I think we can live with the bot failing to add versions, it happens quite a bit -- the alternative would be to update the auto-update config to work for 1.3.1 and then revert that once the version is added, but I think that is more work than is needed, it can just fail to add the missing version.

PR #14243, opened 2024-03-26
swiper-bundle.min.css

lekanshenghuo:


MattIPv4:
As I noted in #14243, if you continued you would be blocked. As such, you are now blocked from the cdnjs org.

